PR #3, opened 2017-07-10
random_walk.py

mattrlam:
Created file to hold other random walk functions.


jgostick:
@mattrlam The names are looking better now, more sensible.  I am able to play with it on my pc, but still have a few questions and suggestions...most about beefing up the docstrings so a user can figure it out a bit quicker.

1.  It is fairly common to put a large docstring on the class itself that outlines all the methods and attributes of the class, and lists references to papers and gives examples of how the code works.  This would be accessed by the user after they've instantiated an object (rw = ps.simulations.RandomWalk(im)), with rw+ctrl-i.  For an example you can check out scipy.spatial.Voronoi.

2.  It's also possible to add a Notes section to the docstring to explain things a bit...for example the ``walk`` function might say that it is only meant for producing single walk for visualization purposes and that the user should use the ``run`` command to perform many walks.  

3. When returning a tuple, consider using a ``namedtuple`` which lets you access the items using attribute names.   Example:

``
from collections import namedtuple
a = namedtuple('msd_result', ('open_space', 'pore_space'))
a.open_space = foo
a.pore_space = bar
``
 
4. ``save_path`` should probably accept a 'path' as an argument, instead of generating a new path?  Is this even possible?  

5.  ``walk_data`` returns None for me

jgostick:
I'm getting this error when I do ``tortuosity``;

Traceback (most recent call last):

  File "<ipython-input-13-a5d11e254065>", line 1, in <module>
    rw.tortuosity()

  File "C:\Users\Jeff\Dropbox\Flash Sync\Code\Git\PoreSpy\porespy\simulations\__RandomWalk__.py", line 328, in tortuosity
    m, m_f = self.run(direction, start_frac, max_steps, stride)

  File "C:\Users\Jeff\Dropbox\Flash Sync\Code\Git\PoreSpy\porespy\simulations\__RandomWalk__.py", line 303, in run
    plt.text(max_steps*0.75, axes.get_ylim()*0.9, p_f[0])

TypeError: can't multiply sequence by non-int of type 'float'

jgostick:
Have you written unit tests for this yet?  They will be able to catch any functions that break or misbehavior, so you should write some fosho.

jgostick:
I have to admit, I'm still having a hard time understanding how this class works.  There are too many functions that do walks in slightly different ways, and it's odd that they get recalculated each time.

I think the class would behave more 'expectedly' if it ran immediately upon initialization. This is how several scipy classes work, such as ``Voronoi`` in the ``spatial`` module:

``` python
rw = RandomWalk(im, nwalkers=500, stride=10, max_steps=5000)
```

If you do it this way, then the properties of the walk could be stored as attributes (i.e. ``nwalkers``).  (It might be handy if ``nwalkers``, ``stride`` and ``max_steps`` were all estimated by the class based on image properties like size and porosity, but this is not necessary)

The returned object should have all results already attached as an attribute.

``` python
rw.path_data
```

All subsequent method calls should look into this data.  

(Note: I know I suggested the ``run`` function, but I'm allowed to change my mind!)

The class should still include methods for visualizing the results that are in ``path_data``.

``` python
rw.plot_msd(direction=None, show_stderr=True)
```

Whether or not standard error is shown should be optional. (Question: If the user puts ``direction`` of [0, 1, 2] is should include all 3, but how to get all 3 PLUS the full distance? )

I guess there should also be a ``get_msd`` function that returns the numerical values as an array, instead of just getting the plot?  (Also, msd is probably better than mean_squared_displacement since its WAY too long.  Again, I'm allowed to change my mind). 

Visualizing a single path is nice, but generating it on demand is a bit confusing since we've already got the data.  This function should reach into the ``path_data`` array and get the path for a given walker:

``` python
rw.show_path(walker=10)
```

I know the ``stride`` higher than 1 will screw it up, but this should be included anyway...users should use ``stride = 1`` if they want an exact path.  The path will still look decent, and might even be clearer...it depends how the path is drawn...as points or connected lines.  In the former case it might look bad, but in the latter case it would be fine.  

As for ``path_data``, might it be cleaner to include the 'index' of the path voxel, rather than the coords?  You can convert back and forth knowing the size of the image using the ravel functions:

``` python
sp.unravel_index(55, dims=[10, 10, 10])
(0, 5, 5)
sp.ravel_multi_index([0, 5, 5], dims=[10, 10, 10])
55
```

There is probably not much chance a user will need to look into this data directly, so obscuring the results as indices instead of coords shouldn't matter.  My feeling is that it actually make the data more clean despite being harder to interpret.  The shape of the data array would be ``nwalkers`` by ``max_steps/stride``.  Interestingly, ``unravel_index`` returns the coordinates in exactly the form that is currently used in ``path_data``.

Maybe there could be a function called ``get_path`` that returns a given path as a coordinate array, instead of asking the user to convert it.  This would partner with ``show_path`` nicely.  

If the user wants to rerun the simulation with a different ``stride`` or ``max_steps``, they should have to reinitialize a new simulation, so the ``clear_walker_data`` function can be removed.  However, if they want to add more walkers, this should be allowed with something like:

``` python
rw.perform_walks(nwalkers=100)
```

This should add new columns to the ``path_data`` arrays, using the ``stride`` and ``max_steps`` values that were given during init, and stored as attributes.  When the class is first initialized, this is the function that should be called.  This is the ONLY function that should do walks.  It should add data to the ``path_data`` array, and all other methods should look into this array.  The user can also set ``nwalkers`` to 1 to get a single walk, then look into the last column of the ``path_data`` array to get the data.  

As a side note, an advanced user can delete the ``path_data`` array, change the ``stride`` and ``max_steps`` attributes on the object, and call ``perform_walks``.  This would in effect clear the entire object and recalculate things.  

The final question is how to get the actual tortuosity from the image.  I have a crazy idea for this...

The challenge is finding the correct start and stop of the slope, so why not plot them all?  Imagine a 2D plot with x-axis showing the starting point and y-axis showing the ending point of where the slope is taken, and the 2D plot shows color for the slope value.  There should be a plateau of near constant values in the middle of the image for all starting points (x) above 1000 and below 4000, and all ending points above 1500 and below 5000...or something like that.  

I'm still not sure how to finish this off, but we can chat about it.

So in summary, the class should have the following methods:

``perform_walks``
``get_msd``
``show_msd``
``show_path``
``get_path``

And the following attributes:

``strides``
``max_steps``
``path_data``

The ``path_data`` array might be cleaner if it were indices instead of coords.

A solid way to calculate ``tortuosity`` still eludes us.

mattrlam:
Working on restructuring the class. I like the idea of storing the path_data on the object, but should I also store the squared displacement data? Or just calculate that when it is needed

jgostick:
somehow Github broke our code!  There were several merge conflicts that did NOT appear in my git client, but only on github.  I tried to resolve them using their web tool but failed it seems.  It looks like the problem is all in the inits...I'll get it fixed asap.

jgostick:
@mattrlam OK, this is sorta working.  The PR passes but the "push" has out of date code...I have no idea what that means.  Anyway, keep plugging away at those upgrades we talked about, and we'll merge it with master soon.

PR #7, opened 2017-08-23
Adding an export module

jgostick:
Created stub code for a ``to_vtk`` function


jgostick:
So is that finished? I recall that @mdrkok abandoned his output to Pallabos temporarily?

mdrkok:
It's done I think, Tom modified the Palabos code to make it faster, I
haven't checked it if the output is still ok but I will,
Matt

jgostick:
@mdrkok when checking, perhaps you could convert your checks in unit tests? 👍

mdrkok:
Will do, just gotta figure out how to do that

PR #10, opened 2017-09-02
Adding 2-point correlation by fft

jgostick:
This function was accidentally added to master, then reverted. The branch was created from the master prior to reverting it so it has the code, but it's not part of the PR (it's part of the master branch), so this PR is only to make fine adjustments to @TomTranter 's great work


jgostick:
This PR is messed up...I've started a new branch with the FFT code on it that is branching from the correct spot on the repo.  I'm going to close this one and delete it.

PR #23, opened 2017-12-05
First attempt at adding dual network extraction

Zohaib-Atiq:
This PR adds a new function to the network_extraction module that extracts both pore and solid phase networks at the same time.


Zohaib-Atiq:
What is codecov/patch and coecov/project Professor?

Zohaib-Atiq:
No idea how to handle codecov/patch errors

jgostick:
Well done, you got pep8 working!  The codecov is another service that runs once the testing build has been successful, and it basically checks how many lines of code were tested.  Since you added about 1000 lines in your function, the overall code 'coverage' went down.  You basically need to write a test that calls your function, then all of your lines would be covered.

jgostick:
I'm thinking about this function that you've been working on.  It's generally a good idea to re-use code...and the problem is that you've cut&pasted the SNOW code into your function.  If we ever find a bug or make a change to SNOW then we might forget to change yours.  It would be better if your function calls SNOW twice, once for pores space, once for solid phase, then adds all the code it needs to blend those two results, as well as extracting the interconnections.  Does that makes sense?

Zohaib-Atiq:
ok I will go through the SNOW code format. I hope i reformat it. Prof. were to look for sample test that you have made for getnet.py ?

Zohaib-Atiq:
Prof. SNOW code will be applied as you have mentioned. There is no modification in snow algorithm. I have changed __getnet__.py function which I cannot run two time separately. The overall procedure will be like applying snow for solid and pore phase and using its outputs as inputs of __dual__ .py to extract dual network. Have i understood your question right?

Zohaib-Atiq:
Prof. I have divided the code in two parts. 1) __getnetv2__py, 2) __dual__py. 
__getnetv2__py is same as __getnet__py but there are some inevitable changes in the code so I have created a new file in case of safety.
__getnetv2__py basically has has 6 inputs instead of 3. New puts are dual, pore_regions and solid_regions.
Default value of solid_regions = None and of dual = False. pore_regions input was introduced in order to align the code for dual extraction, previously we were using im = im * regions to input image mask.  
When dual and solid_regions are at there default value the code will run conventional network extraction algorithm identical to __getnet__py. For dual network extract dual value is set to True and solid_regions are provided to build necessary information for dual network extraction. 
__dual__py have two inputs (net, solid_dt). It will take all information generated from __getnetv2__py and link solid and pore phase together. It is only valid if __getnetv2__py is run by putting dual = True. 
Does this scheme make any sense Prof?

PR #44, opened 2018-08-14
House keeping for V1

jgostick:


jgostick:
why isn't travis running tests on this?

jgostick:
I'm going to merge this into master so that the other branches can pull in these changes (which fix little things that are breaking the build).  We can revisit the idea of cleaning up in a new PR.

PR #59, opened 2018-09-22
Matt changes

jgostick:


mdrkok:
I think its ready, I removed the evtk thing, it became available via pip at some point in the last year

jgostick:
This all looks good!  The code cov is now looking good, so I'll merge.

PR #84, opened 2018-11-08
marching cube area re-implementation for speed improvement  

Zohaib-Atiq:


jgostick:
@Zohaib-Atiq Is this ready for merging into dev?

Zohaib-Atiq:
@jgostick I think yes if you have no further opinion on it.

Zohaib-Atiq:
@TomTranter @jgostick 
run==1 speed comparison
I haven't parrallized the code yet. I have already discussed this with Prof Gostick but we never reached to single point yet but extract_region_area speed can be increased greatly by doing parallel processing. So I can't do the comparison right now. The current code will be of course slower than parallel implementation.

run == 2
Increase the resolution of the image and you will see how much difference your code gives when the image is big. For 100 cubed voxel cube divided into 8 cubes, your alg gives 1344 while mine gives 2407 compared to 2500 theoretical value. I think we discussed this in our previous meeting that at low-resolution surface area will be underestimated so we should avoid using low-resolution image. 

run == 3 
As mentioned above low resolution is a pain to get exact geometrical properties. Perhaps we can use your technique for low-resolution image but it needs to be tested on different shapes structures.

TomTranter:
@Zohaib-Atiq well hopefully the code can serve as a guide on how to parallelize it. Is your code dependent on the level parameter. I don't know if adjusting this based on size would help but running the cube example with 100 side length and changing the level to 0.3 gives pretty close to the right answer.

TomTranter:
@jgostick I think we should merge this in and address these issues in another pr later on

jgostick:
Here is a summary of how I changed the original design of @Zohaib-Atiq:

1.  I have created two functions, ``region_surface_areas`` and ``region_interface_areas``  Their role is self explanatory from the name.  The surface area calculation returns a single array of only surface areas.  The interface area returns an array of areas, but also a second array of conns, so the user knows which regions each area corresponds to.  

2.  I have refactored the actual area calculation to two steps.  First the mesh of the region is found using ``tools.mesh_region``.  This includes the convolution to blur the region.  I have made the strel of the convolution an argument so users can tweak this.  Secondly, the mesh is passed to ``metrics.mesh_surface_area`` which directly calls the function of the same name in skimage.  The one difference is that ``mesh_region`` returns a named tuple for convenience, and our ``mesh_surface_area`` will unpack this if needed, but it also accepts faces and verts, just like the skimage version.

3.  ``combine_regions`` is not actually used...it gets done manually inside each of the original functions, so I will remove this and we can re-add later if needed.  On that note, there are a few functions for dealing with regions that I want to add, so we'll need another PR before moving to V1.0.

TomTranter:
Changes sound good. I assume one of you is looking into why tests are still failing

Zohaib-Atiq:
@jgostick point 1 and 2 no issue I guess. 
point 3 I found `combine_region` to be slow then doing it inside `regions_interfacial_area` but overall this function can be used to mask given regions in an image. If your planning to make new PR then remove it.

jgostick:
@TomTranter I'm working on it and will be done soon.
@Zohaib-Atiq I plan to open another PR with a number of 'region' manipulations in it, such as combine region.  I think it was slow because it was looking at the entire image...the new region ops will hopefully provide a way around this.

TomTranter:
I wouldn't use '../../' try going from porespy.__file__ as root upwards with os.path.join

jgostick:
The problem was actually the LFS file storage thing not being retrieved by travis.  3 hours of my life wasted.

Anyway, this is ready now, but I think @Zohaib-Atiq should run some of his examples on it to make sure he gets the same results.

jgostick:
Nice, only one little catch, when I increase the size of the strel in the `mesh_region` function, the returned region is smoother (as expected), but also significantly bigger!  Can you investigate?  Here is some code to get you started:

```
from skimage.morphology import ball
im = ps.generators.blobs(shape=[200, 200, 200], porosity=0.7, blobiness=1)
regions = ps.filters.snow_partitioning(im)
reg = regions == 10
mesh1 = ps.tools.mesh_region(reg, strel=ball(1))
mesh5 = ps.tools.mesh_region(reg, strel=ball(5))
ps.visualization.show_mesh(mesh1)
ps.visualization.show_mesh(mesh5)
```

Zohaib-Atiq:
That's weird it should show smaller surface when `ball(5)` as compared to ball(1)
```
from skimage.morphology import ball
import porespy as ps
im = ps.generators.lattice_spheres(shape=[51, 51, 51], radius=25)
regions = ps.filters.snow_partitioning(im)
reg = regions == 0
mesh1 = ps.tools.mesh_region(reg, strel=ball(1))
mesh5 = ps.tools.mesh_region(reg, strel=ball(5))
print(ps.metrics.mesh_surface_area(mesh1))  ## 6111.657
print(ps.metrics.mesh_surface_area(mesh5))  ## 5887.3396
```
Is `show_mesh` performing fine? I will check this.

jgostick:
I didn't actually compute the area, just thought that the mesh image was bigger...maybe optical illusion?  Anyway, all looks good then!  I think we can merge it.

PR #101, opened 2018-11-29
Move and improve examples

TomTranter:


TomTranter:
@jgostick I don't know why these tests are failing. They pass locally. Maybe you can help debug as you know more about travis

TomTranter:
Turns out that as openpnm was added as a dependency and porespy is a dependency of openpnm the pip version of porespy was installed and being used instead of the version being tested!!! I added a pip uninstall command to the travis. This is probably also happening on the openpnm branch

PR #103, opened 2018-12-04
Snow n

jgostick:
An attempt to extend snow_dual to snow_n, so that we can extract as many phases as we want.


jgostick:
snow_n is being refactored into more discrete functions on another branch, so this is obsolete.

PR #105, opened 2018-12-04
print statement removed

Zohaib-Atiq:


jgostick:
Maybe you could rename this PR to something related to general bugfixes, and also address the show_mesh bug we found?

Issue #89, opened 2018-11-23
add mode to local_thickness


jgostick:
allow 'fft', and 'dt', and 'mio'

jgostick:
don in #106

Issue #48, opened 2018-09-07
Update regions_to_network function to handle 2D networks correctly


jgostick:
We calculate things like equivalent diameter assuming spheres.  If the image is 2D, we should have an if statement to use circles instead.

jgostick:
The main thing required here is to use some of the 'perimeter' functions in skimage which only work on 2D regions.  These values would be analogous to the surface mesh calculations in 3D.  For instance, throat 'area' would be the perimeter of the two regions minus the perimeter of the combined regions (x2) since the throat area is counted twice (once for each individual region).

PR #106, opened 2018-12-04
General_bug_fixes

Zohaib-Atiq:
This pull request covers small bugs fixes in porespy


jgostick:
any chance of address #48 in this PR?

Zohaib-Atiq:
I think it is related to how we calculate geometrical properties in 2D images. Do you think it is a small fix?

jgostick:
yes, I think you're right...it's not small enough.  Are there any other little bugs we can fix while we're at it though?

jgostick:
Like #89

PR #109, opened 2018-12-11
fixed the porosimetry mio edge issue as well as removed the halo feat…

mdrkok:
Fixes how the mio option in porosimetry handles the edges of the image and also fixes the one voxel feature that was appearing around edges when using the fft option. Two new structuring elements were added to ps.tools, ps_ball and ps_disk. They are cleaner representations of the skimage structuring elements ball and disk which were responsible for the edge bug. @jgostick there is now a minor issue in the mio option when access_limited is True, I know the problem is how the inlets are treated in the padded image but I can't seem to get it to work, should be a quick fix for you

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,7 @@ docs/_build/
 dist/
 
 \.pytest_cache/
+porespy/network_extraction/__snow_dual__.py
+porespy/metrics/.ipynb_checkpoints/__regionprops__-checkpoint.py
+bin/clean
+bin/test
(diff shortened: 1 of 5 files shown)


jgostick:
OK, I'll take a look at the inlets thing, and fix whatever is failing..thanks!

TomTranter:
hey @mdrkok I just noticed glancing through the commits to porespy that you added a couple new strels. I noticed that the skimage ones weren't exactly the same as doing it with a dt when coding the RSA imgen function. The dt is more circular right? Maybe we should change all our circle based imgen functions to use these new ones

mdrkok:
@TomTranter Yah, when I was investigating the issues with the porosimetry  I found that a dt based strel was different then the skimage was. It wasn't just a matter of overall shape but I also found that the way skimage formulates them the sometimes they aren't symmetric for non int inputs like rad=4.25 or something. I think we should try to replace them everywhere, it should be pretty easy, although it may change some outputs slightly which could affect some tests

PR #111, opened 2018-12-13
snow and snow_dual will return im, peaks, dt and regions alongwith net

Zohaib-Atiq:


Zohaib-Atiq:
@jgostick @TomTranter. I have changed `snow` and `snow_dual` output from `dict` to named tuple. Now it will return `im`, `dt`, `regions` and `peaks`. The branch is ready if it ok for you.

Zohaib-Atiq:
Changed the examples and test as well.

jgostick:
I have implemented this in another branch, and this branch didn't quite have it right anyway.

PR #112, opened 2018-12-19
[filter] Consider r_max when looking for peaks

timokoch:
The argument was previously ignored and not passed to the find_peaks function.


Zohaib-Atiq:
@timokoch This bug has been removed in `dev` branch. It will be resolved when we merge the changes in `master`. Thanks for letting us know. 
@jgostick may close this PR if all is fine.

timokoch:
ups didn't know there is a dev branch. Can you add a hint about the branching model that somewhere under contributing/readme?

jgostick:
@timokoch Firstly, thanks very much for making this PR and contributing!  Indeed, we should add contributors guide, and I'll add it to the issues list as high priority.  We'll be merging dev into master very soon (next few weeks) and labeling it version 1.0!  So keep your eye open for the final version.

PR #135, opened 2019-01-19
Generating voxel image from OpenPNM network

ma-sadeghi:
This PR adds the functionality to generate voxel image from OpenPNM network object. A use case is to accurately calculate porosity of a network when there are overlapping pores/throats. Also, it could potentially be used (not yet ready) to accurately calculate the pore surface area to be incorporated in reaction kinetics.

## Automatic usage
You only need to pass the network object and the shape of pores and throats. The method consecutively generates a voxel image from the network, increasing the number of voxels in each iteration until the calculated porosity plateus (tolerance can be passed as `rtol`).
```
import openpnm as op
import porespy as ps
from porespy.network_extraction import generate_voxel_image


net = op.network.Cubic(shape=[5, 5, 5], spacing=1e-3)
geom = op.geometry.StickAndBall(network=net, pores=net.Ps, throats=net.Ts)
im = generate_voxel_image(net, pore_shape="sphere", throat_shape="cylinder")
ps.io.to_vtk(im, path="./net_voxel_image")
```
```
--------------------------------------------
| Generating voxel image from pore network |
--------------------------------------------
Generating pores  : 100%|██████████| 125/125 [00:00<00:00, 1120.51it/s]
Generating throats: 100%|██████████| 300/300 [00:00<00:00, 4173.12it/s]
Generating pores  : 100%|██████████| 125/125 [00:00<00:00, 801.44it/s]
Generating throats: 100%|██████████| 300/300 [00:00<00:00, 3697.09it/s]

Converged at max_dim = 312 voxels.
```
## Manual usage
You override the automatic procedure by passing `max_dim` argument, which is the number of voxels across the maximum dimension of the network.

## Runtime
Currently, the method supports spherical and cubic pores and cylindrical throats. As for the runtime, it's quite fast for regular networks, and slower for random networks (but not terribly slow). Here are some benchmarks:

| Size (# pores) | Cubic (s) | Random (s) |
|----------------|-----------|------------|
| 100            | 0.5       | 1.3        |
| 1,000          | 2         | 4          |
| 10,000         | 10        | 21         |

## Generate images
Here are sample voxel images for two cubic and random networks.
voxel_images


jgostick:
Hey @ma-sadeghi...I know I said to add this to the network_extraction submodule, but now I'm thinking it's more of an IO thing right?

jgostick:
OR...add a helper/wrapper function to IO called ``network_to_im`` or ``openpnm_to_im``, and otherwise keep things as they are?

jgostick:
Also, I wonder if the function should accept a 'resolution' or 'voxel_size' argument, and only do the iteration thing if not given?

ma-sadeghi:
@jgostick I made the changes you suggested. Should be good to go.

PR #166, opened 2019-04-29
Add metadata and license

bryanwweber:
This pull request adds configuration to distribute the `LICENSE` file with the sdist and wheels, as required by the MIT license. It also adds additional project metadata and fixes a few small typos.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ bin/test
 \.idea/
 
 *checkpoint.ipynb
+
+examples/networks/*.vt*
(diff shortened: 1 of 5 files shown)


bryanwweber:
By the way, I didn't add it here, but another good argument to add to `setup()` is the `python_requires` keyword argument with a version specifier for what Python versions this supports.

bryanwweber:
@ma-sadeghi I see that you have `python-requires` for OpenPNM but not here. Should it be the same (`>=3.6`)?

ma-sadeghi:
@jgostick do we have the same restriction for `porespy`? (`Python >= 3.6`)

PR #170, opened 2019-05-15
handle regions given zero label that shouldn't be

TomTranter:


jgostick:
I believe that this was addressed by @Zohaib-Atiq in his recent PR #265, which I just merged.  @Zohaib-Atiq can you confirm, so I can close this and delete the branch?

Zohaib-Atiq:
@jgostick  I discussed this with you in detail I guess. The main problem here is the r_max size. If r_max is large and size of region is small it gets missed during snow_partitioning step. Now when we do masking after segmentation this region is zero. We don't see this effect in SNOW legacy because this region is treated as pore phase there but in SNOW_N zero is considered as dead phase and we get some different isolated regions labeled zero. 
The issue is still open. I suggest we should trim these regions before passing it to snow_partitioning_n.

PR #226, opened 2020-04-03
Customize openpnm_to_im output region(s) to be labeled with pore/throat number

jgostick:


jgostick:
@ma-sadeghi can you investigate this PR?  Might as well merge it into the 1.3 version since it seems like a minor modication.

ma-sadeghi:
@jgostick The `openpnm_to_im` method currently generates a binary image (void -> 1, solid ->0). This PR is an attempt to make it more useful so that pores and throats are identifiable from the image. I did this for Niloo for her ML project. I'm not sure about the API. In its current form, `openpnm_to_im` returns two `im_pores`, which is labelled from 0 to Np-1 (to match OpenPNM's indexing) and with solids labelled as -10 (an arbitrary number), and `im_throats`, laballed from 0 to Nt-1, and with solids labelled as -10. 

What do you think?

ma-sadeghi:
I think this is PR is not that useful, so closing it to not waste our time.

PR #251, opened 2020-05-25
Adding versioneer

jgostick:
Keeping track of versions across branches and PRs, and remembering to update them is driving me nuts.  I'm adding versioneer to the repo in the hopes that it will make life easier.


jgostick:
Damn, I think I know why I removed versioneer!  It drops code coverage by 20%!!  I guess I better find a way to tell codecov to ignore it.

jgostick:
We don't need this now, @ma-sadeghi will have automatic version updates done using github actions.

PR #254, opened 2020-05-28
Add generators.cylinders_porosity

1minus1:
Adding generators.cylinders_porosity and a Jupyter notebook with tests.

This is the first time I've made a PR on any project, so don't mind any mistakes!


jgostick:
Awesome!   I will take a look soon.  BTW, it seems we used up our Actions allotment this month. I will probably ask you to make a small change a 're-push' to trigger the test suite once June rolls around.  In the meantime, I'll skim through the PR.  And thank you very much for this contribution, it's great to know that porespy is worthy of other people's time and effort.

1minus1:
I've made those pep8 compatibility changes in my branch. Let me know when and how to 're-push'.

As for making it a private function, I think it's a good idea not avoid cluttering the namespace, but for consistency's sake with other generators that take porosity inputs (blobs, overlapping_spheres, polydisperse_spheres), it may be better to include porosity as a default input, and offer the ncylinders input as an option for people to keep existing code working. This would likely be a breaking change, but I saw in another thread you were considering those for 2.0. Just my two cents -- you can make the call. and I'll implement it before the re-push.

ma-sadeghi:
Shouldn't the destination be the `dev` branch?

jgostick:
@ma-sadeghi thanks for the reminder!

1minus1:
I see that there are a bunch more errors. I'll clean this up.

jgostick:
Just wait a tick.  I'll deal with the merge conflicts first.

jgostick:
@1minus1, I'm not sure how long it will take for V2...several months probably, so for now, let's keep the function api consistent.

Another thought that occurred to me though was that there should be two hidden functions: ``_cylinders_porosity`` and ``_cylinders_number``.  The  main ``cylinders`` function should analyze the arguments (ie. did the user send ``n_cylinders`` or ``porosity``) and then call the just pass the arguments onto the necessary function. The ``_cylinders_porosity`` function would still call the ``_cylinders_number`` function iteratively.  This would be much easier to maintain in the future.  

As a side note, it also occurred to me that if the user specifies ``n_cylinders`` and ``porosity`` then the ``radius`` becomes the adjustable parameter (i.e. could be set to ``None``).  :-)  Not saying we should do this now, but it's a fun idea.

vasudevanv:
I have been playing around with the side note idea. The way I have it for now is user can specify a radius distribution and number of bins and porosity. The number of cylinders is adjustable to meet the porosity. For now I have the dist and bins hardcoded as I play around with the code. It opens up a lot of possibilities and is fun to play with. 

I have not figured out a nice way to specify the distribution and different combinations of possible parameters and have an API consistent with the existing functions. If this is interesting we could open a issue to continue the discussion.

jgostick:
@vasudevanv In my experience, the easiest way to do 'distributions' is to just call a function several times with different radius and porosity values, then add the images together.  We actually have a function called ``polydisperse_spheres`` that does this by calling ``overlapping_spheres``.  It takes a scipy.stats distribution as an argument.  It's not perfect, but maybe could give you an idea.

vasudevanv:
@jgostick Thanks! I will take a look at that function.

1minus1:
There are some clever ways to implement your sidenote. I'll take a swing at it soon.

1minus1:
I wasn't aware that the cylinders generated by generators.cylinders have rounded ends. Is this intentional? It doesn't bother me, but might be worth a note in the docstring. 

This code:

`im2 = ps.generators.cylinders([200,200,200], 20,10,0,90,40)`

gives this image
image

Anyway, this is a happy accident for the purpose of this discussion, because it made writing cylinders_n_poro much easier. I create cylinders of radius 1, calculated the edt, then threshold this to create an image with a new fiber diameter (finding the value using a bisection algorithm on np.unique(edt.edt(im)). This approach gives cylinders with rounded ends just like generators.cylinders and is really fast!

I'll include this in another PR once we come to agreement on an approach.


```
def cylinders_n_poro(shape ,ncylinders: int, porosity:float,  phi_max: float =0, theta_max: float = 90, length: float = None):
    im = ps.generators.cylinders(shape,1,ncylinders,phi_max,theta_max,length)
    dt = edt.edt(im)

    current_porosity = ps.metrics.porosity(im)

    if porosity > current_porosity:
        raise Exception('This combination of n_cylinders and porosity cannot be met with this set of inputs (requested porosity too high, or ncylinders too low).')
    
    vals = np.unique(dt)[1:]
    i=0
    stepsize = int(np.floor(vals.size /2))
    porofromindex = lambda index: ps.metrics.porosity(dt >= vals[index])

    # find the right threshold using a bisection algorithm

    print('starting dt iteration')
    while current_porosity > porosity:
        current_porosity = porofromindex(i)
        if current_porosity < porosity:
            if stepsize == 1:
                im = (dt >= vals[i])
                print('Resulting radius is {}'.format(1+vals[i]))
                return im
            else: 
                i=i-stepsize
                current_porosity = porofromindex(i)
                stepsize = int(stepsize/2)
        else:            
            i += stepsize
```

jgostick:
Hi @1minus1 I'm sorry that I drop the ball on this one.  By looking at the failed builds, it seems like the failures were due to pep8 warnings, meaning you didn't adhere to python style guides.  Looking at the logs I see you were inconsistent with spaces and commas.

ma-sadeghi:
Hi @1minus1, thanks for this PR, and sorry about the long wait. I took the liberty of refactoring your code. I also added a unit test to make sure everything is OK. 

@jgostick Can you review this PR? Thanks!

porespy/generators/__imgen__.py
```diff
@@ -974,30 +975,157 @@ def cylinders(shape: List[int], radius: int, ncylinders: int,
...
+        on the randomly chosen center point of the cylinder.
+    max_iter : scalar
+        The number of fractional fiber insertions used to target the requested
+        porosity. By default a value of 3 is used (and this is typically
+        effective in getting very close to the targeted porosity), but a
+        greater number can be input to improve the achieved porosity.
+    return_fiber_number : bool
+        Determines whether the function will return the number of fibers
+        along with the image
+
```
jgostick, line 1061: verbose is missing from this list.  HOWEVER, I have been thinking about making this a package wide setting, so I think we should remove it from this function for now, and come up with general solution.

porespy/generators/__imgen__.py
```diff
@@ -974,30 +975,157 @@ def cylinders(shape: List[int], radius: int, ncylinders: int,
...
+    for i in range(1, max_iter):
+        fractions.append(fractions[i - 1] + (max_iter - i) ** 2 * subdif)
+
+    im = np.ones(shape, dtype=bool)
+    for frac in tqdm(fractions, disable=not verbose, file=sys.stdout,
+                     desc="Adding fibers"):
+        n_fibers_total = n_pixels_to_add / vol_fiber
+        n_fibers = int(np.ceil(frac * n_fibers_total) - n_fibers_added)
+        if n_fibers > 0:
+            im = im & ps.generators.cylinders(
```
jgostick: I'm inclined to think that we should make the current 'cylinders' function a hidden method (_cylinders) and make this new one the official 'cylinders' method, and add only the `porosity` to the func call.  If `porosity=None` and `n_cylinders=None` are the defaults, then the user can specify one or the other, and so both new and old behaviors can be obtained.

porespy/generators/__imgen__.py
```diff
@@ -974,30 +975,157 @@ def cylinders(shape: List[int], radius: int, ncylinders: int,
...
+        will exist. If a scalar value is given it will be interpreted as the
+        Euclidean distance between the two ends of the cylinder.  Note that
+        one or both of the ends *may* still lie outside the domain, depending
+        on the randomly chosen center point of the cylinder.
+    max_iter : scalar
+        The number of fractional fiber insertions used to target the requested
+        porosity. By default a value of 3 is used (and this is typically
+        effective in getting very close to the targeted porosity), but a
+        greater number can be input to improve the achieved porosity.
+    return_fiber_number : bool
```
jgostick, line 1058: This seems like a strangely specific flag to put into the func call, and I really hate the idea of returning a tuple sometime and an image others.  I assume that scott (1minus1) has a need for this, but maybe he was just trying to be 'symmetrical' with the original cylinders function, where the user knows the number of cylinders by definition?  I wonder if there is another way to count the cylinders?

porespy/generators/__imgen__.py
```diff
@@ -974,30 +975,157 @@ def cylinders(shape: List[int], radius: int, ncylinders: int,
...
+            im = im & ps.generators.cylinders(
+                shape, radius, n_fibers, phi_max, theta_max, length, verbose=False
+            )
+        n_fibers_added += n_fibers
+        # Update parameters for next iteration
+        porosity = ps.metrics.porosity(im)
+        vol_added = get_num_pixels(porosity)
+        vol_fiber = vol_added / n_fibers_added
+
+    if verbose:
```
jgostick: Regarding my above comment about the module-wide settings, we should also consider using the logger for these printed messages, so they can be turned off too, but setting loglevel very high.  In any case, remove verbose from here, so it always prints.  This is currently the normal porespy behavior.

porespy/generators/__imgen__.py
ma-sadeghi: Good point, done

porespy/generators/__imgen__.py
ma-sadeghi: I agree, @1minus1 do you need to know the number of cylinders in your use case? Would it be OK if we just return an image?

porespy/generators/__imgen__.py
ma-sadeghi: We need to somehow be able to control `_cylinders`, since it's being called by `cylinders`, and nested `tqdm`s - while generally supported - are not currently supported in Spyder (there's an open issue since Aug 2018, so I guess they're not working on it).

I can keep the `verbose` argument in `_cylinders`, which is a hidden method anyway, and remove it from our public `cylinders`.

jgostick: I don't think tqdm is really needed in _cylinders...the slowest step is the distance transform to dilate the fibers, so just remove it

ma-sadeghi:
I'll merge this PR for now, if we decide to also return the number of cylinders, we can do it later.

porespy/generators/__imgen__.py
1minus1: My use of returning the number of fibers was primarily for testing purposes, so I think it's fine to remove it.

ma-sadeghi: @1minus1 Great. Btw, we merged your PR with a slight change. We made the previous `cylinders` method private as `_cylinders` and renamed your `cylinders_porosity` to `cylinders`. Also, we combined their arguments, so that you can either pass porosity or number of cylinders as input.

PR #266, opened 2020-06-20
Updating JOSS DOI (and badge)

arfon:
:wave: @jgostick - I'm not sure if you have a reason for using the Zenodo DOI here but it's not actually the correct DOI for the JOSS paper. 

This PR updates the preferred citation DOI and adds the JOSS badge to your README :-)


jgostick:
I didn't realize there was a difference, but thanks so much for this.

Issue #172, opened 2019-06-02
add mode convolve to fftmorphology


jgostick:
This would allow me to use fftconvolve (which fftmorphology) calls without having to remember where it is in scipy.  But beyond laziness, it basically calls the function with no frills, so does the convolve and passes it through.

jgostick:
I guess if the strel was a gaussian kernal it could do gaussian blurs.

jgostick:
I started a PR for this, thinking it'd be easy, but it was not easy.  I am closing both.

PR #270, opened 2020-07-08
Add convolve option to fftmorphology

jgostick:
This fixes issues #172


jgostick:
this is not worth the trouble.

PR #280, opened 2020-07-27
Add export to OpenFOAM

jgostick:


ma-sadeghi:
Hi @magnaou, can you post here a minimal working OpenFOAM script that uses the output of `to_openfoam`? Thanks

magnaou:
I just added an example with c++ OpenFoam example

jgostick:
@magnaou We need you to take another pass over this PR.  I changed the code a bit as follows:
* deleted the openfoam class and renamed ``save`` to ``to_openfoam``
* removed ``to_openfoam`` wrapper from the __funcs__ file, and imported it directly from the __openfoam__ file
* improved the path handling
* remove the zoom feature (users can do that themselves outside the func)
* renamed the scale to ``voxel_size`` to be consistent with other funcs
* renamed ``labels`` to ``label_boundaries``

It now successfully writes the blockMeshDict file, sorta...

What we still you to **explain how to use this file**!  I have actually managed to get openfoam on my win10 machine using WSL2, but don't know how to use this mesh file.  A notebook example would be great!  You mention above that you'd added an example but I don't see it anywhere.

jgostick:
The problem is:
* 2D images get converted to 3D (using np.atleast_3d)
* If label_boundaries == False, it all works
* If True however, the boundary addition fails.
* This is only an issue with X by Y by 1 images.
* It'd be nice to support this if possible.

ma-sadeghi:
@jgostick @magnaou any update on this?

ma-sadeghi:
I copied this branch to our private repo so we can delete it from here since there hasn't been any progress on this for a long time.

PR #282, opened 2020-08-09
Return diameter for local_thickness calculation

akin-aroge:
Local thickness should return diameter not radius

### DIFF
--- a/porespy/filters/__funcs__.py
+++ b/porespy/filters/__funcs__.py
@@ -1105,7 +1105,7 @@ def local_thickness(im, sizes=25, mode='hybrid'):
 
     """
     im_new = porosimetry(im=im, sizes=sizes, access_limited=False, mode=mode)
-    return im_new
+    return im_new * 2
 
 
 def porosimetry(im, sizes=25, inlets=None, access_limited=True,


jgostick:
Hi @akin-aroge Thanks very much for this well-formatted patch.  I appreciate your point that it's unclear whether radius or diameter should be used.  I don't think changing the output of a function like this is backwards compatible, since other people will depend on receive radii values.  I do agree that the doc string should be more clear that radii values are returned, instead of 'pore size'.  

BTW, if I were to change something significantly, I would rather change the name of the function to something more descriptive like ``local_diameter``.  We use local thickness since this is what imageJ uses.

ma-sadeghi:
@jgostick I assume this can be closed then?

OhmFF:
This may be too late, but this question has been in my mind for a long time. Could you please clarify it? @jgostick @ma-sadeghi 

For example, if we have a 1-voxel pore, the local thickness would return the value of 1, right?

If it is converted to pore size, it means the radius of 1 voxel (aka 2 voxels in diameter), which does not make sense.

This is confusing when plotting `ps.metrics.pore_size_distribution` scaling with the voxel resolution and multiplying by 2 to make it into diameter for comparison with other characterization methods, such as mercury intrusion porosimetry.

PR #316, opened 2020-10-14
Fix a bug in add_boundary_regions that led to unlabeled pores

### LINKED ISSUE #315: Bug in SNOW algorithm when adding boundary regions
Also relevant #238

ma-sadeghi:
Fixes #315

When adding boundary faces, sometimes there are some unlabeled pores, meaning that suppose we have `boundary_faces = ["left", "right"]`, and we're extracting a dual phase network, labeled `void`, and `solid`, now if we count up the total number of pores using `net.pores(["left", "right", "void", "solid"])`, in principle we should get the same number as `net.Np`, but we don't.


ma-sadeghi:
@jgostick Could you please review this PR? I think I might have finally found the bug. 

The bug was that when incrementing padded regions (i.e. the boundary regions), we should have only incremented non-background voxels since padded background voxels must always remain as background, i.e. labeled 0. So, for instance, let's say your background is touching the edge of the image at some locations. Now, when we pad the image, we have some 0s in the padded regions. Suppose `regions.max()` is 25, adding this to the padded regions makes those 0s, 25s, whereas 25 is an internal region!

What I did to fix it was to only increment the non-boundary padded regions, i.e. leave out those that are 0, and increment the rest by `regions.max()`. Does that make sense?

Btw, I refactored the `add_boundary_regions` method to simplify the logic (at least that was what I hoped for!).

porespy/networks/__funcs__.py
```diff
@@ -77,85 +77,110 @@ def add_boundary_regions(regions=None, faces=['front', 'back', 'left',
...
+    # Convert pad-induced corners to 0
+    _zero_corners(regions, pad_width=3, faces=faces)
+
+    # Make labels contiguous
+    regions = make_contiguous(regions)
 
     return regions
 
 
+def _zero_corners(im, pad_width: int, faces: list):
```
jgostick: I like this function, but I think the api should match the numpy.pad, so instead of faces, should be axes, like:
_zero_corners(im, pad_width=[[0, 2], [0, 2]]) would pad 2 rows at the end of the x and y axes, but not at the start, while _zero_corners(im, pad_width=[[3, 3], [0, 2]]) would zero both ends of the x axis 3 deep, but 2 tall.  And _zero_corners(im, pad_width=[[3, 3], [0, 0]]) would do nothing, since the y size of the pad is 0.

porespy/networks/__funcs__.py
```diff
@@ -77,85 +77,110 @@ def add_boundary_regions(regions=None, faces=['front', 'back', 'left',
...
+    r"""
+    Fills corners of a padded-image with 0, given a pad width and list of
+    faces such as ["left", "front"] at which the image has been padded.
+
+    Example
```
jgostick: You should make this example actual code by putting >>> for each line...then pytest will test it.

porespy/metrics/__funcs__.py
```diff
@@ -735,23 +738,25 @@ def region_surface_areas(regions, voxel_size=1, strel=None):
...
+    print('Finding surface area of each region', flush=True)
     im = regions.copy()
     # Get 'slices' into im for each pore region
     slices = spim.find_objects(im)
     # Initialize arrays
     Ps = np.arange(1, np.amax(im)+1)
     sa = np.zeros_like(Ps, dtype=float)
     # Start extracting marching cube area from im
-    for i in tqdm(Ps):
+    tqdm_Ps = tqdm(Ps, file=sys.stdout)
```
jgostick: If we're going to change the tqdm behavior, it should be:
with tqdm(iterator) as pbar:
   pbar.update()
   <code here>
This way the pbar always closes properly even if the code errors.  Here is an example: https://github.com/PMEAL/porespy/blob/e82c22e3923e6f0b89db85588f5745cc3c1f4f6d/porespy/filters/__funcs__.py#L1305

porespy/tools/__funcs__.py
```diff
@@ -5,7 +5,7 @@
 from edt import edt
 from collections import namedtuple
 from skimage.morphology import ball, disk
-from skimage.measure import marching_cubes_lewiner
+from skimage.measure import marching_cubes
```
jgostick: let's keep things backwards compatible, so do
try:
    from skimage.measure import marching_cubes
except ImportError:
    from skimage.measure import marching_cubes_lewiner as marching_cubes

porespy/networks/__funcs__.py
jgostick: also, it might be ok to make this a public function?  Put it in tools, and don't forget to add it to the init file.

porespy/metrics/__funcs__.py
jgostick: Maybe you can find-all and update the tqdm usage everywhere while you're at it?

jgostick (changes requested):

porespy/metrics/__funcs__.py
ma-sadeghi: Yep, I did. I just realized that the issue with tqdm is "not flushing the sys.stdout" rather than tqdm object not getting closed. I tested it and adding `file=sys.stdout` fixes the flush problem. I changed it everywhere.

porespy/networks/__funcs__.py
ma-sadeghi: Done

porespy/networks/__funcs__.py
ma-sadeghi: Done

porespy/tools/__funcs__.py
ma-sadeghi: Done

Issue #319, opened 2020-10-21
regionprops_3D


hsafari68:
When tried to get the properties on a 3D carbonate image labeled by snow for a 600 by 600 by 150 volume, the function was very very slow and I got the following error every time and ofcourse a couple of other errors. 
But when I tried the regionprops method in skimage, results were returned pretty fast without error.

Error:
QhullError: QH6013 qhull input error: input is less than 3-dimensional since all points have the same x coordinate    0

While executing:  | qhull i Qt
Options selected for Qhull 2019.1.r 2019/06/21:
  run-id 236552816  incidence  Qtriangulate  _pre-merge  _zero-centrum
  _max-width  4  Error-roundoff 5.5e-15  _one-merge 3.9e-14
  _near-inside 1.9e-13  Visible-distance 1.1e-14  U-max-coplanar 1.1e-14
  Width-outside 2.2e-14  _wide-facet 6.7e-14  _maxoutside 4.4e-14

hsafari68:
It works on smaller subvolumes, but crashes on larger ones

jgostick:
Quite honestly, this function is a bit obsolete now.  I wrote it back before scikit-image had proper 3d support in this function.  Is there something specific you're looking for that is not in scikit-image?

jgostick:
also, that particular error means that the region is flat, like 2d...qhull does not like that.

hsafari68:
> 
> 
> Quite honestly, this function is a bit obsolete now. I wrote it back before scikit-image had proper 3d support in this function. Is there something specific you're looking for that is not in scikit-image?

I actually used regionprops_3D for 2D images and it worked quite well. It gave me a couple of other properties than skimage regionprops function like sphericity, but that's not a big deal. 
What I liked was the props_to_DataFrame method that I used to convert all the returned proeprties to a data frame in Python, but unfortunately that also seem not to be working giving me the same errors as regionprops_3D did. These functions seemed to be quite handy in 2D images, but I assume I have to write a couple of lines of codes to handle these calculations for my case unless you would want to update these functions in the porespy package.

jgostick:
ok, I'm on it.  But first, are you sure that you just don't have an anomalous image?  Does it work as desired with other 3D images?

hsafari68:
I would really appreciate it if you do because I have found porespy and openpnm very powerful packages in digital rock physics.
[...]

PR #320, opened 2020-10-22
Fixing up regionprops_3d

jgostick:
This function has needed some love for a while.  I finally looked into how skimage does their delayed dispatch on the regionprops function.  They create an RegionProps object for each region, which has attributes that you access, THEN they calculate the property.  So, I subclassed this and added our own to the mix.  

The actual attrs that we compute should probably be rethought, but that would be backward incompatible, so I guess we should leave it for now?


jgostick:
Fixes #319

PR #331, opened 2020-12-09
Save an image as a paraview state file and open it in Paraview

ma-sadeghi:


ma-sadeghi:
@hfathian Everything looks great, you just need to add two unit tests for `to_paraview` and `open_paraview` methods. You can find some examples here: https://github.com/PMEAL/porespy/blob/dev/test/unit/test_io.py

ma-sadeghi:
@jgostick @hfathian Paraview devs just pushed a fix to take care of the MacOS issue. Should be on conda soon (a few days possibly).

PR #340, opened 2021-01-14
Gravity packing

jgostick:
A simple function for inserting spheres into an image at the lowest accessible point mimicking gravity since they all settle to the bottom of the image


jgostick:
The image below shows large particles added with a clearance of 10, then small particles added sinking to the bottom, then medium ones added which get clogged at the top:

sample

ma-sadeghi (approved):

jgostick:
Hey @ma-sadeghi ... the fixed you added doesn't work.  The self.surface_area call is so that it runs and the next line won't fail.  I shouldn't return anything.  Maybe it should be _ = self.surface_area, but _ will just be None.

ma-sadeghi:
> Hey @ma-sadeghi ... the fixed you added doesn't work. The self.surface_area call is so that it runs and the next line won't fail. I shouldn't return anything. Maybe it should be _ = self.surface_area, but _ will just be None.

I see. Quick question: if it doesn't do anything, what's the point of the whole `if` statement?

ma-sadeghi:
> > Hey @ma-sadeghi ... the fixed you added doesn't work. The self.surface_area call is so that it runs and the next line won't fail. I shouldn't return anything. Maybe it should be _ = self.surface_area, but _ will just be None.
> 
> I see. Quick question: if it doesn't do anything, what's the point of the whole `if` statement?

Nevermind, I see the point now. I'm wondering if there's a better way to achieve this (not that it's too important anyway).

PR #341, opened 2021-01-16
Adding image-based invasion percolation algorithm

jgostick:


jgostick:
The two remain errors are fixed on the gravity packing branch, so will leave until that gets merged into dev

jgostick:
The coverage for numba functions is not counted.  I added quite a few for the ibip method so it appears like the coverage of the diff is very low.

porespy/filters/__ibip__.py
```diff
@@ -0,0 +1,334 @@
...
+                pt = np.vstack(np.where(bd))
+                # scratch.fill(True)
+                # scratch *= bd
+                scratch = np.copy(bd)
+                temp = _insert_disks_at_points(im=scratch, coords=pt,
+                                               r=t, v=1, smooth=False)
+            # Reduce to only the 'new' boundary
+            edge = temp*(dt > 0)
+            if ~np.any(edge):
+                print('\nNo more accessible invasion sites found...exiting')
```
ma-sadeghi, line 105: We should probably use `loguru`'s "info" instead of `print` statements.

porespy/filters/__ibip__.py
```diff
@@ -0,0 +1,334 @@
...
+            pt = np.where(edge*dt_thresh)  # Keep as tuple for later use
+            inv = _insert_disks_at_points(im=inv, coords=np.vstack(pt),
+                                          r=r_max, v=step, smooth=True)
+            if return_sizes:
+                sizes = _insert_disks_at_points(im=sizes, coords=np.vstack(pt),
+                                                r=r_max, v=r_max, smooth=True)
+            bd[pt] = True  # Update boundary image with newly invaded points
+            dt[pt] = 0
+            if step == (max_iters - 1):  # If max_iters reached, end loop
+                print('\nMaximum number of iterations reached...exiting')
```
ma-sadeghi, line 121: This also needs to be `loguru.info`

ma-sadeghi (changes requested):

PR #344, opened 2021-01-23
Adding submodule for imagej wrappers

jgostick:


ma-sadeghi:
In a discussion with @hfathian, we figured out that since `pyimagej` doesn't work in headless mode, tests will likely fail (I wonder how Windows tests passed). So we have two options:

1. Get rid of `test_imagej.py` altogether
2. Keep `test_imagej.py`, but manually ignore it in the Actions

None are ideal though. @jgostick What do you think? maybe a third option?

ma-sadeghi:
Being pursued in two other PRs #386, #390.

PR #349, opened 2021-01-25
Faster builds

jgostick:


jgostick:
@ma-sadeghi is there a reason you never created a PR for this?

ma-sadeghi:
> @ma-sadeghi is there a reason you never created a PR for this?

I don't think it was significantly faster, I'll get back to this later this week to see what can be done.

PR #358, opened 2021-02-05
Fix the check whether the image is a boolean in snow partitioning

colfrog:
Since im.dtype returns a string, checking whether im.dtype is bool will always be false, so snow_partitioning will always create a full copy of the image in binary format, regardless of the image being binary.

This causes the function to always take more memory than is required. This patch fixes the check in order to achieve expected behaviour.

### DIFF
--- a/porespy/filters/__funcs__.py
+++ b/porespy/filters/__funcs__.py
@@ -249,7 +249,7 @@ def snow_partitioning(im, dt=None, r_max=4, sigma=0.4, return_all=False,
     print("-" * 60)
     print("Beginning SNOW Algorithm")
     im_shape = np.array(im.shape)
-    if im.dtype is not bool:
+    if im.dtype is not np.dtype("bool"):
         print("Converting supplied image (im) to boolean")
         im = im > 0
     if dt is None:


jgostick:
Hi @colfrog, thanks for this.  Nice catch.  I will address the error which is because a code snippet in the docstring is now outputting a message that pytest is not expecting.

jgostick:
Hi @colfrog I edited your fork directly.  The problem was actually that the comparison to a string was also not right.  The correct way to do is ``im.dtype != bool``, so compare to the bool class, but use the comparison operator rather than the `is`

jgostick:
Well this is annoying...the ``im.dtype != bool`` check works just fine on my pc.  The image that is used in that function is surely a boolean so this logic should not be called.

colfrog:
Strange, were there cases where the comparison to a string didn't work? The absolute right way to do it is to compare with an `is`, against the same type of object as returned by `im.dtype`. I've just updated the commit to do exactly that.

jgostick:
No, I don't think ``is`` is correct because ``is`` checks instances, while the ``im.dtype`` returns a numpy type object not a boolean instance.  The comparison operator is implemented on the dtype object, so it compares to bool.  something is fishy here.  Maybe the image returned from blobs is actually not a a bool?  I'll dig into it more.

jgostick:
i have refactored the snow algs so this is no longer part of the workflow

PR #359, opened 2021-02-05
Allow passing the binary image to find_peaks to reduce memory load

colfrog:
My development machine has too little memory, and my testing dataset was too big, so I looked for extra memory to save.

I left it as is for the call to chunked_func with find_peaks, because it's simpler that way.

This patch shouldn't change the existing API.


jgostick:
I'm going to close this pr because find peaks should only accept at a distance transform.  It's up to the user to obtain it and massage it as needed (blur, etc), THEN they can find the peaks.

PR #367, opened 2021-02-14
Adding settings to top-level of package

### LINKED ISSUE #201: Add quiet option to functions to turn off progress bar and messages

### LINKED ISSUE #210: Add a settings dict to the module?
It would be handy to add ``settings`` to the top level of the module.  The first use case would be to turn on/off messages output by functions, or the progress bar.  Perhaps it could even specify which progress bar to use, since the terminal and the notebook ones look much different.  Could also specify colormaps, number of cores to use for chunked functions, etc.

### LINKED ISSUE #354: Ensure all tqdm progress bars are used with a context manager
This prevents them breaking when the function ends unexpectedly.

jgostick:
closes #354.  I altered the tqdm progress bars to check the ``show_progress`` setting, so converted them all to context manager style while I was at it.

closes #210
closes #201


jgostick:
closes #354.  I altered the tqdm progress bars to check the ``show_progress`` setting, so converted them all to context manager style while I was at it.

ma-sadeghi:
Nice!

How about replacing the explicit context manager with the implicit one? This way, there's one level less indentation, which makes the code more readable, plus there's no need to manually updating the progress bar.

```python
from tqdm import tqdm
from time import sleep

# Explicit context manager
with tqdm(range(5)) as pbar:
    for i in range(5):
        sleep(0.2)
        if i == 2:
            break
        pbar.update()
print("BREAK")

# Implicit context manager
for i in tqdm(range(5)):
    sleep(0.2)
    if i == 2:
        break
print("BREAK")

# No context manager (the old way)
pbar = tqdm(range(5))
for i in range(5):
    sleep(0.2)
    if i == 2:
        break
    pbar.update()
print("BREAK")
```
# Explicit context manager
```bash
 40%|█████████████████████████████████▌                                                  | 2/5 [00:00<00:00,  3.17it/s]
BREAK
```
# Implicit context manager
```bash
 40%|█████████████████████████████████▌                                                  | 2/5 [00:00<00:00,  3.14it/s]
BREAK
```
# No context manager (the old way)
```bash
 40%|█████████████████████████████████▌                                                  | 2/5 [00:00<00:00,  4.73it/s]BREAK
 40%|█████████████████████████████████▌                                                  | 2/5 [00:00<00:00,  3.12it/s]
```

jgostick:
The explicit context manager is required so that the printout doesn't break if the loop is interrupted.  Both of the implicit approaches above will break.

ma-sadeghi:
I don't think it does. See the console printout in my previous comment (it contains a `break` when `i == 2`), it only broke for the manual progress bar.

jgostick:
but what about when a user hits ctrl-c, or an unexpected error occurs?  When this happens, each new instance of tqdm put a newline character after each update so the progress bar ends up on a new line on each iteration.  I *think* the context manager approach is the only way to prevent this, in my experience.

jgostick:
Btw, the middle option above is the OLD way.  AFAIK, we've never used the 3rd option

ma-sadeghi:
I just checked w/ my Spyder, and the implicit context manager works fine with Ctrl+C as well.

I think we currently use a mix of both. I found 3 in filters and 1 in generators (the manual progress bar).

PR #379, opened 2021-02-17
Adding showdoc func

jgostick:


jgostick:
Hey @ma-sadeghi, I'd like to merge this asap so that @madeline-am can start using it in her notebook examples.  Can you give it a review?  I changed the behavior so that is uses try/except on the import and just returns a text string if the module is not install, so you always get something returned.

ma-sadeghi (approved):
Looks good!

PR #408, opened 2021-03-10
Accept masks as inlets throughout the package

### LINKED ISSUE #353: Enforce all funcs to only accept boolean masks for borders and boundaries
For instance, trim_nonpercolating_paths accepts an inlet_axis and outlet_axis.  This is not always clear how it works, so I'd like to always and only accept boolean masks for things like this.

### LINKED ISSUE #404: Move get_border to generators module, and rename to border
This is technically a 'generator', so it'd be better placed in generators I think.

### LINKED ISSUE #405: Change behavior of border (formerly get_border) so that edges means corners in 2D
At the moment edges is interpreted to mean faces, which is more literal.  But functions that call this method would work better if edges and corners were equivalent in 2D.

### LINKED ISSUE #407: Added faces function for generating images for inlets and outlets
Since we're moving to masks as indicators for inlets, we'll need a function to generate them.

jgostick:
fixes #407
fixes #405 
fixes #404
fixes #353


jgostick:
Merge message should read:
```Added borders and faces to generators #new Enhanced add_boundary_regions function to accept pad_width instead of face names #api```

PR #423, opened 2021-03-20
New noise

jgostick:
Add a ``fractal_noise`` function to ``generators``.  It uses ``pyfastnoisesimd`` to generate fractal noise.


ma-sadeghi:
@jgostick The only issue is that it's not available on `conda-forge`

jgostick:
I added ``pip install requirements/pip_requirements.txt`` to the ci yamls.  There will probably always be a few packages that are not on conda yet, so I have emptied out the pip_requirements file, except for 1 package that is optional but should be tested.

PR #433, opened 2021-03-29
Convenience wrapper for mayavi.mlab.volume_slice

hfathian:


ma-sadeghi:
I copied this branch to our private repo so we can delete it from here, since there hasn't been much progress on this PR.

PR #508, opened 2021-04-21
Improve phase fractions

### LINKED ISSUE #506: tqdm settings not used in geometrical_tortuosity and ibip

### LINKED ISSUE #507: metrics.phase_fractions requires phase labels to contain no gaps
# Issue with gaps in the labels
## Current output
```
q
Out[23]: array([ 0,  0,  1,  1, 12,  1, 12, 12, 12, 12])

ps.metrics.phase_fraction(q)
Out[24]: array([0.2, 0.3, 0. , 0. , 0. , 0. , 0. , 0. , 0. , 0. , 0. , 0. , 0.5])
```

## Expected
```
q
Out[23]: array([ 0,  0,  1,  1, 12,  1, 12, 12, 12, 12])

ps.metrics.phase_fraction(q)
Out[24]: array([0.2, 0.3, 0.5])
```

## Or even better
```
q
Out[38]: array([ 0,  0,  1,  1, 12,  1, 12, 12, 12, 12])

ps.metrics.phase_fraction(q)
Out[44]: {0: 0.2, 1: 0.3, 12: 0.5}
```

# Issue that the method doesn't work with floats
## Current output
```
x
Out[15]: array([ 0.5,  0.5,  1.5,  1.5, 12. ,  1.5, 12. , 12. , 12. , 12. ])

import porespy as ps

ps.metrics.phase_fraction(x)
Traceback (most recent call last):

  File "<ipython-input-17-f62def6fb91a>", line 1, in <module>
    ps.metrics.phase_fraction(x)

  File "c:\users\amin\code\porespy\porespy\metrics\__funcs__.py", line 812, in phase_fraction
    raise Exception('Image must contain integer values for each phase')

Exception: Image must contain integer values for each phase
```
## Expected
```
x
Out[46]: array([ 0.5,  0.5,  1.5,  1.5, 12. ,  1.5, 12. , 12. , 12. , 12. ])

ps.metrics.phase_fraction(x)
Out[47]: {0.5: 0.2, 1.5: 0.3, 12.0: 0.5}
```

ma-sadeghi:
Fixes #507 
Fixes #506


jgostick:
I think this function would be best if it returned a dict with phase_alias as the dict keys, no?

ma-sadeghi:
> I think this function would be best if it returned a dict with phase_alias as the dict keys, no?

Do you mean the user would have to pass `phase_alias` to `phase_fraction` as well?

jgostick:
I was thinking that optionally they could pass ``phase_alias = {'solid': 0, 'void': 1, 'binder': 2}``, which would return ``{'solid': 0.44, 'void': 0.22, 'binder': 0.34}``.  If they don't pass anything, then it would return ``{0: 0.44, 1: 0.22, 2: 0.32}``.  In fact, I think this is probably essential, since it's the only way to handle missing values, like what if there is no 1 in the image?  The current way would return ``[0.22, 0.78]``...or would it return ``[0.22, 0, 0.78]``?  And what about float values?  How do you associate a value like 1.5 to a position in a list?

ma-sadeghi:
It used to return `[0.22, 0, 0.78]`. I changed it so instead of keys being `np.linspace(0, im.max())`, they are `np.unique(im)`. So, if there's no 1, it returns: `{0: 0.22, 2: 0.78}`.

For floats, let's say the image contains 30% 1s and 70% 1.5s, currently it returns: `{1: 0.3, 1.5: 0.7}`.

PR #509, opened 2021-04-21
Sort imports using isort package

ma-sadeghi:


ma-sadeghi:
@jgostick do you agree with this change?

jgostick:
you ran isort on the code, but did you also add it to the actions or is this a one time thing?

ma-sadeghi:
It's a one time thing, but I can add it as an Action if you think is important.

jgostick:
No it's ok, i was just curious.  I think we could add it as an action when we also add yapf.  For now however, I think we should close this PR.  Let's do this as the very last thing before releasing 2.0.  This will cause too many conflicts with various things that are still being worked on, including the notebooks.

PR #525, opened 2021-04-25
Rewrite trim_saddle_points

### LINKED ISSUE #510: Fix "Maximum number of iterations reached" warning

jgostick:
Fixes #510
This function has been a pain for a while with the max_iters thing.  I have re-wrote it.   It can still in principle hit a max_iter, but I've added a few other stopping criteria so that it should almost never happen.


jgostick:
I tried removing the need for max_iter, but its necessary when ridge points are found, neither peaks or saddles.  It seems a default of 20 is enough to prevent the warning from appearing.  The main thing is that the method now has some different logic that prevents this from being triggered accidentally and unavoidably as was the case before.

PR #550, opened 2021-07-15
Expose resequence labels

jgostick:
The make_contiguous function was SUPER slow, so we moved to using the skimage 'rankdata' function, but this seem to use a lot of memory, so this PR exposes a previously hidden function called resequence_labels that @Zohaib-Atiq wrote for his parallelization project.  It's numba jitted, and parallelized.  It does not have the fully functionality of the make_contiguous function, so let's hope it works.

### DIFF
--- a/porespy/filters/_snows.py
+++ b/porespy/filters/_snows.py
@@ -10,7 +10,7 @@ from skimage.segmentation import watershed
 from skimage.morphology import ball, disk, square, cube
 from porespy.tools import _check_for_singleton_axes
 from porespy.tools import extend_slice
-from porespy.filters import chunked_func
+from porespy.filters import chunked_func, resequence_labels
 from porespy import settings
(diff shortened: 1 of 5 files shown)


jgostick:
This function is fast (50 ms on 10 million elements) but has some problems, primarily that the order of the labels may change.  This is ok for resequencing a watershed, but is a problem for invasion sequences for instance.  The new version of make_contiguous that uses the skimage label_sequential (instead of scipy.stats rankdata) takes about 500 ms, so it 5x faster than the current version but still 10x slower than resequence_labels. I guess this extra time is required for the assurance that the label order does not change.  I'm going to close this pr now and we can go with the new version of make_contiguous.  If that turns out to be too slow, we can revisit this function.

ma-sadeghi:
Copied to our private repo.

PR #551, opened 2021-07-15
IBIP algorithm

Ni2M:
This PR handles the IBIP algorithm on CPU and GPU.


ma-sadeghi (changes requested):
Good job @Ni2M, could you please change rename the gpu version of the methods you added to `_gpu` rather than `_g`? I think it's more descriptive. Thanks!

Ni2M:
For sure, yes the naming _gpu is more descriptive. Thanks.

Ni2M:
The PR is still under process. We are yet to determine final changes on how we are adding the ibip_gpu (whether to keep importing cupy,cupyx and dependency to cupy or alternatives).

jgostick:
Yes, just get everything passing now, then we can finalize the naming and api during the devs meeting on Monday

ma-sadeghi:
> The PR is still under process. We are yet to determine final changes on how we are adding the ibip_gpu (whether to keep importing cupy,cupyx and dependency to cupy or alternatives).

Just a few notes to be discussed during the dev meeting:
- I don't think GitHub CI machines have GPUs, I'm not sure if `cupy` can even be installed (maybe it can?)
- Since we can't assume that everyone has access to a dedicated Nvidia graphics card (`cupy`'s AMD support is experimental), I think it's best if we list `cupy` as an "optional" dependency, like `paraview`.

PR #565, opened 2021-07-21
Re-enable imagej tests

hfathian:


ma-sadeghi:
I had to temporarily disable imagej tests, also excluded the entire imagej module from coverage. It's hard to get `pyimagej` to work on CIs (issues like OS-dependency and not working with Python 3.9, etc.).

PR #608, opened 2021-12-23
Drop python 3.6 and 3.7 support per NEP29

jamesobutler:
The dataclasses module was introduced in Python 3.7 (see https://docs.python.org/3/library/dataclasses.html) and therefore not available to use in Python 3.6. The dataclasses module began to be used in `porespy` starting in https://github.com/PMEAL/porespy/commit/4630768d1160e9db5dcfcd2e40c754b1e6509863.

Python 3.6 has also gone end-of-life as of today (https://endoflife.date/python) which is a reason not to make changes to continue to use this codebase on Python 3.6.

This change to drop Python 3.6 support matches what the classifiers are already indicating for this repo. 
https://github.com/PMEAL/porespy/blob/ec9791e63db6e6a1281e364f4d2ea5d3796f70c9/setup.py#L34-L39

I will warn that because `porespy` has quite a large number of dependencies, that not specifying some version limits on dependencies can lead to this package becoming broken once the first dependency decides it is time to drop some Python version. For example `numpy` 1.22.0 will be releasing soon (currently at [1.22.0rc03](https://github.com/numpy/numpy/releases/tag/v1.22.0rc3)) and it is dropping Python 3.7 support per [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html). Therefore `porespy` will need to become documented as Python >= 3.8 as well.

### DIFF
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: 3.7
+        python-version: 3.8
 
     - name: Set env variables
       run: |
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: 3.8
+        python-version: '3.8'
 
     - name: Set env variables
       run: |
--- a/.github/workflows/ci-macos.yml
+++ b/.github/workflows/ci-macos.yml
@@ -16,8 +16,8 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.8]
-        operating-system: [macos-latest]
+        python-version: '3.8'
+        operating-system: macos-latest
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/ci-macos.yml
+++ b/.github/workflows/ci-macos.yml
@@ -16,8 +16,8 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: '3.8'
-        operating-system: macos-latest
+        python-version: ['3.8']
+        operating-system: [macos-latest]
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/ci-min.yml
+++ b/.github/workflows/ci-min.yml
@@ -15,8 +15,8 @@ jobs:
     strategy:
       max-parallel: 1
       matrix:
-        python-version: '3.8'
-        operating-system: ubuntu-latest
+        python-version: ['3.8']
+        operating-system: [ubuntu-latest]
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/ci-min.yml
+++ b/.github/workflows/ci-min.yml
@@ -15,8 +15,8 @@ jobs:
     strategy:
       max-parallel: 1
       matrix:
-        python-version: [3.8]
-        operating-system: [ubuntu-latest]
+        python-version: '3.8'
+        operating-system: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.8, 3.9]
         operating-system: [ubuntu-latest]
 
     steps:
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -16,7 +16,8 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: ['3.8', '3.9', '3.10']
+        # Add '3.10' to the list once #611 is addressed
+        python-version: ['3.8', '3.9']
         operating-system: [ubuntu-latest]
 
     steps:
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.8, 3.9, 3.10]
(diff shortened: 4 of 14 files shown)


jgostick:
Thank you very much for this PR.  I was actually surprized that we were using dataclasses in porespy...I'd forgotten about adding it I guess.  However, instead of this version contraint I am more inclinded to remove the dataclass decorator from the Settings class, since we don't use any of the dataclass functionality.  Of course, we still have to worry about the numpy version support as you say.  

Versioning in python is getting more problematic all the time.  I really don't want to force users to create virtualenvs and to remember to activate it each time they fire up jupyter or spyder.  Our aim is for anyone to be able to install porespy in their base anaconda install, if they wish.

ma-sadeghi:
@jgostick If numpy is dropping support for 3.7, maybe we should too?

jamesobutler:
If you don't update `porespy` based on the minimum python version of all your dependences, your usage of `numpy`, `scipy` and all other scientific packages following NEP29, is going to break functionality here in your package. Running `porespy` on Python 3.6 is going to run into issues when it uses latest `numpy` and latest `numpy` uses functionality only present in Python 3.7+. 

Expecting users to then follow up their `porespy` install by downgrading their versions of `numpy` is going to be difficult. With appropriate versioning set, `porespy` development can continue, but users using unsupported versions of Python by the various dependencies can instead use an older version of `porespy` that will then get the older dependencies.

jgostick:
Our official stance on this is to always support the latest version of numpy.  I "think" that is written in the versioning.md file.

jamesobutler:
Ok that sounds good! Since the latest version of `numpy` [1.21.5](https://github.com/numpy/numpy/releases/tag/v1.21.5) requires Python >=3.7 (this requirement has been since version [1.20.0](https://github.com/numpy/numpy/releases/tag/v1.20.0) released January 30th 2021), does that mean this PR is in line with your desire to only support Python 3.7 and later as well?

> `PoreSpy` depends on several other packages widely known as the [Scipy Stack](https://www.scipy.org/stackspec.html).  It is our policy to always support the latest version of all these packages and their dependencies.

^ From versioning.md, to support the latest versions of these packages that will require following their minimum required versions of Python which includes Python 3.7 and later as of today. In a few days it will be Python 3.8 and later when `numpy` goes from rc3 to stable release of 1.22.0.

jamesobutler:
Currently this [Scipy Stack](https://www.scipy.org/stackspec.html) link in the versioning.md document results in 404.

I do know Scipy adheres to [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html) 

> This NEP recommends that all projects across the Scientific Python ecosystem adopt a common “time window-based” policy for support of Python and NumPy versions. Standardizing a recommendation for project support of minimum Python and NumPy versions will improve downstream project planning.

> This NEP is being put forward by maintainers of Matplotlib, scikit-learn, IPython, Jupyter, yt, SciPy, NumPy, and scikit-image.

The implementation section states

> We suggest that all projects adopt the following language into their development guidelines:
> 
> This project supports:
> 
> - All minor versions of Python released 42 months prior to the project, and at minimum the two latest minor versions.
> - All minor versions of numpy released in the 24 months prior to the project, and at minimum the last three minor versions.
> In setup.py, the python_requires variable should be set to the minimum supported version of Python. All supported minor versions of Python should be in the test matrix and have binary artifacts built for the release.
> 
> Minimum Python and NumPy version support should be adjusted upward on every major and minor release, but never on a patch release.

[source](https://numpy.org/neps/nep-0029-deprecation_policy.html#implementation)

ma-sadeghi:
@jamesobutler Then please make any necessary changes to this PR so it conforms to NEP29.

jamesobutler:
@ma-sadeghi Ok, here are some dates below related to NEP29. Since Dec 26th has passed since I originally opened this PR, there will need to be a change for `porespy` to only support Python 3.8+ in its next release. 

I’m currently away from my office, but in a couple of days I can update various locations in this repo including your doc files for this Python 3.8+ change. Or if you would like to do that sooner than a couple days from now, feel free to merge this PR and then submit that change as a follow up.

Past Dates:
> - On Jun 23, 2020 drop support for Python 3.6 (initially released on Dec 23, 2016)
> - On Dec 22, 2021 drop support for NumPy 1.18 (initially released on Dec 22, 2019)
> - On Dec 26, 2021 drop support for Python 3.7 (initially released on Jun 27, 2018)

Future:
> - On Apr 14, 2023 drop support for Python 3.8 (initially released on Oct 14, 2019)

ma-sadeghi:
@jamesobutler No rush :) Btw, really appreciate this PR. Thanks so much!

jamesobutler:
@ma-sadeghi I've pushed updates to reflect support for Python 3.8 and later per NEP29. Support dates are listed below.
```
On Jan 07, 2020 drop support for NumPy 1.14 (initially released on Jan 06, 2018)
On Jun 23, 2020 drop support for Python 3.6 (initially released on Dec 23, 2016)
On Jul 23, 2020 drop support for NumPy 1.15 (initially released on Jul 23, 2018)
On Jan 13, 2021 drop support for NumPy 1.16 (initially released on Jan 13, 2019)
On Jul 26, 2021 drop support for NumPy 1.17 (initially released on Jul 26, 2019)
On Dec 22, 2021 drop support for NumPy 1.18 (initially released on Dec 22, 2019)
On Dec 26, 2021 drop support for Python 3.7 (initially released on Jun 27, 2018)
On Jun 21, 2022 drop support for NumPy 1.19 (initially released on Jun 20, 2020)
On Jan 31, 2023 drop support for NumPy 1.20 (initially released on Jan 30, 2021)
On Apr 14, 2023 drop support for Python 3.8 (initially released on Oct 14, 2019)
On Jun 20, 2023 drop support for NumPy 1.21 (initially released on Jun 19, 2021)
```

ma-sadeghi:
Looks great! We'll merge this PR once tests all pass. Thank you so much for your contribution! :tada:

jamesobutler:
Per NEP29

> Minimum Python and NumPy version support should be adjusted upward on every major and minor release, but never on a patch release.

@ma-sadeghi Therefore these commits should only be released as part of a `porespy` 2.1.0 or 3.0.0.

ma-sadeghi:
@jamesobutler Yeah, I think we're going to release v2.1 soon, we actually have a [milestone](https://github.com/PMEAL/porespy/milestone/2), of which 6 issues have not yet been addressed :cry:

PR #613, opened 2022-01-05
Added edge option

jgostick:
adding an optional argument to trim functions to remove things touching the edges/surfaces of the images too.

### DIFF
--- a/porespy/filters/_funcs.py
+++ b/porespy/filters/_funcs.py
@@ -213,7 +213,7 @@ def distance_transform_lin(im, axis=0, mode="both"):
     return f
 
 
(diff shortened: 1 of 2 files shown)


jgostick:
Damn, more weirdness...two versions fail, but the rest pass.  This is also happening in openpnm.

ma-sadeghi:
That's strange...I'm sure I've skipped those two tests on dev. Anyway, it seems to have worked fine. The tests on dev are passing.

jgostick:
Yes, I noticed that the "push" tests were failing but the same tests on "PR" were ok, so I rolled the dice and merged it ;-)

PR #659, opened 2022-03-18
Snow2 maintenence

### LINKED ISSUE #317: Improving unit tests for network extraction functions
Some of our methods are not properly being tested by our unit tests. I only skimmed through `test_network_extraction.py`, and it seems that we're not properly testing network extraction. From what I see, most of the time, we're only checking the shape of the output, whereas at the very least, I think we should check the values.

### LINKED ISSUE #573: Why does snow2 expect `phases` indices to be contiguous?
Why not make it more general, so that `phases` can contain any arbitrary sequence of labels, e.g., `[1, 5, 10, 12]`.

### LINKED ISSUE #638: SNOW2 comprehensive tests take almost half of the pytest runtime
It could be made faster by using a smaller image. @jgostick if you think it's worth it, could you please take a look?

jgostick:
fixes #573
fixes #317
fixes #638

Commit message:
Added ability to pass custom peaks to snow algorithms #enh
Phase labels no longer need to be contiguous in snow functions or snow2 extraction #enh
Added many quantitative unit tests for snow functions #maint
Using much small images in snow tests to improve testing speed #maint
Phase labels of 0 can now be passed to snow2, if phase_alias argument provided #enh
Minor adjustments/improvements to various trim peaks functions #maint


ma-sadeghi:
@jgostick Thanks! What happens if we pass the following alias? `{0: "phase1", 1: "phase2", 4: "phase3"}`
Does it ignore `phase1` altogether?

PR #664, opened 2022-03-24
Fixing region_volumes method

### LINKED ISSUE #663: metrics region_volumes gives the same or negative values for different regions
The problem is when choosing 'voxel' method, the returned volumes of regions is the same for all pore regions (this can be fixed by changing:
https://github.com/PMEAL/porespy/blob/76e0c3132e69d07eeb115d872302ad12203a287a/porespy/metrics/_meshtools.py#L49
to:
```python
vols[i]=region.sum()
```
Another problem is when choosing 'marching cube' method, the returned volumes of regions are all negative. I will be checking the root cause and updating.

example of the problem:
```python
import matplotlib.pyplot as plt
import numpy as np
import porespy as ps
np.random.seed(10)
im = ps.generators.blobs(shape=[50,50,50])
snow = ps.filters.snow_partitioning(im)
regions = snow.regions
vols = ps.metrics.region_volumes(regions=regions)
print(vols[:5])
#[ -17.74722219 -103.5931484   -18.03024671 -231.43682392 -689.57165046]
vols = ps.metrics.region_volumes(regions=regions, mode='voxel')
print(vols[:5])
#[7474405. 7474405. 7474405. 7474405. 7474405.]
```

Ni2M:
This PR closes #663.


Ni2M:
The problem with 'voxel' mode is fixed. About the 'marching cube' mode after investigating the code I could not find any bug. I think the marching cube volume approximation is correct, but  just with a negative sign.
See following example for a closed spherical shape:
```python
from skimage.morphology import ball
import matplotlib.pyplot as plt
import porespy as ps
region=ball(10)
plt.imshow(region[:,:,5])
vol_march=ps.metrics.region_volumes(regions=region)
vol_vox=ps.metrics.region_volumes(region, mode='voxel')
print('voxel mode, marching cube mode volume=', vol_vox, vol_march)
# voxel mode, marching cube mode volume= [4169.] [-4102.28678846]
```
The volume calculations (negative values) happens in Trimesh package which is hard coded in Trimesh\Triangles.py file. In porespy, we can change the results after the fact:
https://github.com/PMEAL/porespy/blob/76e0c3132e69d07eeb115d872302ad12203a287a/porespy/metrics/_meshtools.py#L73
to:
```python
        vol = -1*m.volume
````

**Comment by Ni2M** [comment]
Now the code works properly. Rerun the first example from [here](https://github.com/PMEAL/porespy/issues/663#issue-1179671652) on this branch:
```python
import matplotlib.pyplot as plt
import numpy as np
import porespy as ps
np.random.seed(10)
im = ps.generators.blobs(shape=[50,50,50])
snow = ps.filters.snow_partitioning(im)
regions = snow.regions
vols = ps.metrics.region_volumes(regions=regions)
print(vols[:5])
#[ 17.74722219  99.73476231  18.15524676 167.29447398 683.4471061 ]
vols = ps.metrics.region_volumes(regions=regions, mode='voxel')
print(vols[:5])
#[ 24. 130.  32. 200. 768.]
```

jgostick:
- [x] change -1*x to abs(s)
- [x] add unit tests for voxel and mesh methods

PR #673, opened 2022-03-31
Diffusive conductance using AI

ma-sadeghi:


Ni2M:
Summary of changes:
- Add try except for tensorflow imports
- change the name to AI_diffusive_size_factor (diffusive part)
- make conns a mandatory argument
- build the model before calling the AI method (using create_model() and load_weights)
- pass the model and training data object (instead of path)
- Always predict using tensorwise (delete tensorwise Boolean argument)
- After applying region_to_net, apply the AI_diffusive_size_factor method and assign the results to net['AI_diff_size_factor']
- Update example
- (created a downloadable link for all files required to run the example)

Summary of checks:
- The method works for both 1 pair of regions and segmented regions
- The runtime for create_model() to load the weights on is 0.84 sec.

porespy/networks/_AI_diffusive_size_factor.py
```diff
@@ -0,0 +1,405 @@
...
+    -------
+    t_conns : array
+        An Nt by 2 addat containing the throats' connections in the segmented image.
+
+    '''
+    struc_elem = ball
+    slices = spim.find_objects(im)
+    Ps = np.arange(1, np.amax(im)+1)
+    t_conns = []
+    for i in tqdm(Ps, file=sys.stdout):
```
jgostick: This line should be ``tqdm(Ps, **settings.tqdm)``. assuming you've called ``from porespy import settings``.  Also, you could add ``d = 'Finding neighboring regions'`` and the call to tqdm would be:  ``tqdm(Ps, desc=d, **settings.tqdm)``

porespy/networks/_AI_diffusive_size_factor.py
```diff
@@ -0,0 +1,405 @@
...
+
+    '''
+    if g_train is None:
+        raise ValueError("Training ground truth data must be given\
+                         to be used for normalizing the test data")
+    its = -1
+    pairs = np.empty((len(throat_conns), 64, 64, 64), dtype='int32')
+    diff_size_factor = []
+    zm_ratios = []
+    for cn in throat_conns:
```
jgostick: Maybe add tqdm here too?

Ni2M:
summary of final changes applied:
- updated tqdm in AI diff size factor file
- finalized tutorial example and reference example
- moved imports of tensorflow and scikit-learn into the functions (similar to ibip_gpu) to prevent pytest errors
- added ignore pytest for AI_diff_size_factor example files to prevent pytest errors

PR #685, opened 2022-04-20
Fix diffusive conductance using AI example notebooks

Ni2M:
The build doc issue occurs because it can't find the files to run the example build. The files are downloadable and the user needs to save them locally to use the examples. This PR creates an ignore build error for sphinx build documentations.


Ni2M:
created a [per-notebook setting](https://nbsphinx.readthedocs.io/en/0.4.1/never-execute.html#Explicitly-Dis-/Enabling-Notebook-Execution) as [JSON metadata](https://nbformat.readthedocs.io/en/latest/format_description.html) to disable executing AI example build. Note that per-notebook settings take precedence over the settings in conf.py. If this per-notebook setting doesn't work, we can change `nbsphinx_execute = 'auto'` in conf.py so that notebooks with no outputs are executed and notebooks with at least one output([pre-executed notebooks](https://nbsphinx.readthedocs.io/en/0.4.1/pre-executed.html)) are not.

ma-sadeghi:
`nbsphinx_execute = 'auto'` works, but has a serious downside: notebooks are no longer guaranteed to work w/ the dev version, since they're not being rerun. We want examples to always work using the dev version. That way, if a change recently made to dev breaks one of the examples, we won't get notified.

ma-sadeghi:
I think you can download files from internet in a jupyter notebook. If you create a repo and put the required files in it, you just need to clone that repo in the notebook using:

```
!clone REPO_URL
```

Ni2M:
> `nbsphinx_execute = 'auto'` works, but has a serious downside: notebooks are no longer guaranteed to work w/ the dev version, since they're not being rerun. We want examples to always work using the dev version. That way, if a change recently made to dev breaks one of the examples, we won't get notified.

That is a good point. For that reason I was also hesitated to change the conf.py for all notebooks and only implemented per-notebook settings for the AI examples. Thank you for the suggestion about the repo idea. However, we could not upload the files on a repo using regular git, because of their large size. git LFS would be an alternative solution, although it may make things more difficult for the user (I think they have to install git LFS [before cloning](https://www.atlassian.com/git/tutorials/git-lfs#clone-respository)). Currently, we have included the files as downloadable dropbox files so that the user can download and use them with the provided example.

ma-sadeghi (approved):

Ni2M:
I think this PR is ready to merge. Thanks.

PR #701, opened 2022-06-03
Add files via upload

PascalRuzzante:
Hi Jeff and Amin,

I've made a few changes to the two-point correlation function and the associated radial profile helper function. Initially inspired by needing to scale the normalized probability by the porosity, in order to fit with the Torquato definition, but I ended up making a number of changes (improvements, I think!). 

1. Scaling the probability by porosity
2. Added a new result attribute for the normalized probability (same as previous probability).
3. Allowed passing in of number of desired bins, or a pre-specified bin array (bin edges)
4. If passing in number of desired bins, I started the counting from r = 0 instead of r = bin_size
5. Changed the deprecated fftpack to fft module. This allowed the use of rfftn and irfftn, used for fourier transforms on real numbers, and results in a significant reduction in computation time.
6. Make use of scipy.fft set_workers function to enable parallel processing for further speed increase. 5 and 6 enabled a 2x reduction in computation time in my tests.
7. Added some of the functionality of _parse_histogram to radial profile so that user can access bin centers, bin edges, etc. I left cdf out as it seem to me the probability is already somewhat cumulative (my statistics definitions are pretty rusty, correct me if I'm wrong here)
9. Allowed user to specify voxel size.

I've never submitted a pull request before, so I'm hoping this work, but let me know if I missed anything.

Best,
Pascal

### DIFF
--- a/porespy/metrics/_funcs.py
+++ b/porespy/metrics/_funcs.py
@@ -1,7 +1,8 @@
 import numpy as np
 import scipy.ndimage as spim
 import scipy.spatial as sptl
-from scipy import fftpack as sp_ft
+import multiprocessing
+from scipy import fft as sp_ft
 from skimage.measure import regionprops
 from deprecated import deprecated
 from porespy.tools import extend_slice
@@ -601,7 +602,7 @@ def two_point_correlation_bf(im, spacing=10):
     return tpcf
 
 
-def _radial_profile(autocorr, r_max, nbins=100):
+def _radial_profile(autocorr, bins, pf=None, voxel_size=1):
     r"""
     Helper functions to calculate the radial profile of the autocorrelation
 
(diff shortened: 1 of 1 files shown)


jgostick:
This looks AMAZING, and we're so grateful for your time!  I will review the code soon, but it looks like you've done all the PR stuff correctly.  One thing that jumps out at me is lack of unit tests.  I can give you a tutorial in writing and adding tests after our next class if you have time, or during office hours next week.

ma-sadeghi:
Hi @PascalRuzzante, thanks so much for your contribution!

PascalRuzzante:
No problem! I have to proctor a make-up quiz on Tuesday from 11:30 to
12:30, but could come by after that, if you're not busy? Otherwise, I can
of course come on Friday for office hours. I've never done a unit test
before, so you may be starting from square 1 with me. Is it written into
the code itself, or are they separate, as part of github?

jgostick:
You can see what's involved by checking out the existing ones here:  https://github.com/PMEAL/porespy/blob/be4a129bc28891622d7ee9e813d2e27f6dd0e1bf/test/unit/test_metrics.py#L50.   Basically you just add a function test check to ensure the code gives the expected output, and add it to the 'tests' folder, then pytest finds it.

PR #703, opened 2022-06-14
Updating two point correlation

PascalRuzzante:
I've created a new pull request  because something weird was happening with viewing diffs between the one I submitted previously. For the record, I am copying over the changes I put in that PR comment.

1. Scaling the probability by porosity
2. Added a new result attribute for the scaled_probability
3. Allowed passing in of number of desired bins, or a pre-specified bin array (bin edges), which can be generated from np.arange, linspace or logspace, for example.
4. If passing in number of desired bins, I started the counting from r = 0 instead of r = bin_size so the first point corresponds to r=0
5. Changed the deprecated fftpack to fft module. This allowed the use of rfftn and irfftn, used for fourier transforms on real numbers, and results in a significant reduction in computation time.
6. Make use of scipy.fft set_workers function to enable parallel processing for further speed increase. 5 and 6 enabled a 2x reduction in computation time in my tests. **Updated this to use ps.setting.ncores**
7. Added some of the functionality of _parse_histogram to radial profile so that user can access bin centers, bin edges, etc. I left cdf out. **Because of the way the tpcf works, I couldn't figure out a way to pass it to np.histogram first, and therefore couldn't pass in a density=True argument. Because of this, I made a small change to the _parse histogram method by adding in a keyword argument that only runs the added portion if set to False (defaulted to True)**
8. Allowed user to specify voxel size. **This is applied after the fact, so bins should be supplied in terms of voxels.**
9. **Further increased the speed of the function by extracting the for loop in the _radial_profile function to a separate function, modified slightly so that I could run it with the @njit(parallel=True)**
10. **Added a unit test to check the initial and final values of the scaled probability**

I haven't checked the overall speed increase over the current version, but I believe it is significant, which was important for my usage, in which I am calling it iteratively many times. This will also depend on the machine used, and the number of bins used (surprisingly this is fairly significant).


jgostick:
This all looks really great.  I'm going to merge it into our ``dev`` then we can tweak as needed.

PR #754, opened 2022-11-15
Add mode to imagej_wrapper and imagej_plugin

mkaguer:
Added mode to pyimagej functions because 'headless=False' is now deprecated.


jgostick:
I'm going to close this because we need to rethink the imagej integration

PR #796, opened 2023-03-08
Add function to compute diffusive size factor using dns

jgostick:


Ni2M:
The PR is not ready to merge yet. I'm still investigating the backend code and my archives to make sure the correct function is used here.

Ni2M:
I have finalized my changes. I think this is ready to merge. Thanks.

ma-sadeghi (commented):
@Ni2M `tqdm` has produced lots of cruft in `diffusive_size_factor_DNS` notebook, could you please silent it? You can do that by modifying porespy's settings.

ma-sadeghi (changes requested):
@Ni2M `tqdm` has produced lots of cruft in `diffusive_size_factor_DNS` notebook, could you please silent it? You can do that by modifying porespy's settings.

Ni2M:
Thank you Amin for your review, for sure. I just wanted to let you know I removed my finalization comment, as adding a new tutorial to this PR is in progress. Thanks.

jgostick:
An example was accidentally committed directly to dev.  This commit was 'reverted', which I had a vague recollection can cause some problems.  After some digging I found [this answer](https://stackoverflow.com/a/71828272) which points out that you must cherry-pick the commit(s) from dev onto the desired branch, not merge them, otherwise they are ignored when merged back onto dev.  Anyway, I have done as described in this question, which corresponds to my vague memory of when this happened before, so let's hope it works.

PR #807, opened 2023-03-15
Handling dicts better in Results object

### LINKED ISSUE #777: Printing of the Results class should ignore dicts
When printing the Results object returned by ``snow2`` it displays the entire network dict which makes the printout useless.  It should probably handle these more cleverly.

jgostick:
Fixes #777


jgostick:
Improved output of Results object when large dicts present #bug

PR #810, opened 2023-03-17
Fix table describing result attributes

rickyfann:


ma-sadeghi (changes requested):
@rickyfann Please add `ps.visualization.set_mpl_style()` to the cell with library imports, so graphs look nicer.

ma-sadeghi (commented):
Looks good!

PR #827, opened 2023-04-28
subtracting the throat cross-sectional areas from the pore surface ar…

avlmachado:
I made a modification to the regions_to_network function to subtract the cross-sectional area of throats from the surface area of pores.

A correct prediction of the solid-fluid area is necessary for mass transfer processes, such as catalysis, heterogeneous reactions, adsorption, among others.

### DIFF
--- a/porespy/networks/_getnet.py
+++ b/porespy/networks/_getnet.py
@@ -252,6 +252,8 @@ def regions_to_network(regions, phases=None, voxel_size=1, accuracy='standard'):
                                                 voxel_size=voxel_size)
         A = interface_area.area
         net['throat.cross_sectional_area'] = A
+        np.add.at(net['pore.surface_area'],net['throat.conns'][:, 0], -A)
+        np.add.at(net['pore.surface_area'],net['throat.conns'][:, 1], -A)
         net['throat.equivalent_diameter'] = (4*A/np.pi)**(1/2)
     else:
         net['pore.volume'] = np.copy(p_volume)*(voxel_size**ND)


jgostick:
Hi @avlmachado, thanks very much for this.  In general we have been trying to do a little as possible in porespy and leave the details up to the end user in OpenPNM, but in this case I think your suggestion is reasonable.  I have a few concerns though

- What about N phase extractions?  What is the meaning of pore surface area if there are multiple phases?  
- If we do go ahead with this, should we perhaps call it something else, like `'pore.phase1_phase2_interfacial_area'`, and keep the `'pore.surface_area'` as is?  This way things are backward compatible for people who are expecting `'pore.surface'` area to mean the entire area of the pore.

ma-sadeghi:
@jgostick Maybe pore.surface_area should be called pore.region_surface_area, to be consistent with pore.region_volume? And not have pore.surface_area at all, and let openpnm import function deal with it (possibly subtract throat area as @avlmachado suggested)?

avlmachado:
Hi @jgostick , I had forgotten that the algorithm has an option to extract more than 2 phases. I think renaming the current function to pore.region_surface_area as @ma-sadeghi suggested seems adequate, as well as updating the variable description.
Maybe have something like pore.phase1_phase2_interfacial_area only for the snow_partitioning algorithm and not snow_partitioning_n?

jgostick:
So after talking wtih @Ni2M , she pointed out that the code already removes this surface area. It gets done on [this line](https://github.com/PMEAL/porespy/blob/fada48f4e1b79c4d6b1d6b36e48a435865c1fc63/porespy/networks/_getnet.py#LL198C17-L198C36).

avlmachado:
Hi @jgostick, this line only removes the area for the 'standard' accuracy method, for high accuracy calculation this is not done.

PR #828, opened 2023-05-02
Throat cross-sectional area calculation is wrong when `accuracy="high"`

ghost:
Hello everyone,
There is an error in calculating the throat cross-sectional area using the high accuracy method.

The surface area used as input has to be without multiplying by the voxel size.

The following test demonstrates the problem:

```
import porespy as ps
import numpy as np
import openpnm as op
np.random.seed(13)

resolution = 5e-7

im = ps.generators.overlapping_spheres([100, 100, 100], r=7, porosity=0.7)

snow_output = ps.networks.snow2(im,
                   voxel_size=resolution,
                   boundary_width=3,
                   accuracy='high', # high
                   parallelization=None)

pn = op.io.network_from_porespy(snow_output.network)

A = pn['throat.cross_sectional_area']
print(A[1:10])
```
When subtracting the surface area of merged regions from the isolated region surface areas, they need to be in the same unit.

### DIFF
--- a/porespy/networks/_getnet.py
+++ b/porespy/networks/_getnet.py
@@ -248,7 +248,7 @@ def regions_to_network(regions, phases=None, voxel_size=1, accuracy='standard'):
         net['pore.volume'] = region_volumes(regions=im, mode='marching_cubes')
         areas = region_surface_areas(regions=im, voxel_size=voxel_size)
         net['pore.surface_area'] = areas
-        interface_area = region_interface_areas(regions=im, areas=areas,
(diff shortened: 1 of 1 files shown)


ma-sadeghi:
@jgostick I haven't looked into it, but if the OP’s description is accurate, seems serious.

jgostick:
Sadly, I need to close the PR because I didn't deal with it in a timely manner and now the simple fix conflicts too much with the target branch. I will take the suggested code and apply it in a different PR.

PR #839, opened 2023-05-23
Added log message to the begining and ending of every functions

### LINKED ISSUE #838: Add logger.info to beginning and end of each function
The logger outputs the time it writes, so this can be used to get a rough idea of how long a function takes by setting loglevel to 2, then just looking at the before/after of the logger message.  I'm running some jupyter notebooks and curious to know how long various functions take for large iamges, so find myself wishing I could just set the loglevel to info to get this.

Daniel-olaO:
close #838

### DIFF
--- a/porespy/__init__.py
+++ b/porespy/__init__.py
@@ -39,10 +39,13 @@ def _setup_logger_rich():
     import logging
     from rich.logging import RichHandler
 
+    logger = logging.getLogger(__name__)
+    logger.info('_start of _setup_logger_rich')
     FORMAT = "%(message)s"
     logging.basicConfig(
         format=FORMAT, datefmt="[%X]", handlers=[RichHandler(rich_tracebacks=True)]
     )
+    logger.info('end of _setup_logger_rich')
 
 
 _setup_logger_rich()
--- a/porespy/filters/_funcs.py
+++ b/porespy/filters/_funcs.py
@@ -52,24 +52,37 @@ logger = logging.getLogger(__name__)
 @deprecated("The ibip function will be moved to the"
             + " ``simulations`` module in a future version")
 def ibip(**kwargs):
+    frame = insp.currentframe()
+    function_name = insp.getframeinfo(frame).function
+    logger.info("start of" + function_name)
     r"""
     This function has been moved to the ``simulations`` module, please use that.
     """
     from porespy.simulations import ibip
+    logger.info("end of" + function_name)
     return ibip(**kwargs)
+    
 
 
 @deprecated("The ibip_gpu function will be moved to the"
             + " ``simulations`` module in a future version")
 def ibip_gpu(**kwargs):
+    frame = insp.currentframe()
+    function_name = insp.getframeinfo(frame).function
+    logger.info("start of" + function_name)
     r"""
     This function has been moved to the ``simulations`` module, please use that.
     """
     from porespy.simulations import ibip_gpu
+    logger.info("end of" + function_name)
     return ibip_gpu(**kwargs)
 
 
+
 def find_trapped_regions(seq, outlets=None, bins=25, return_mask=True):
+    frame = insp.currentframe()
+    function_name = insp.getframeinfo(frame).function
+    logger.info("start of" + function_name)
     r"""
     Find the trapped regions given an invasion sequence image
 
@@ -138,6 +151,7 @@ def find_trapped_regions(seq, outlets=None, bins=25, return_mask=True):
         # of zero needs to be removed, if it's in keep.
         keep = np.setdiff1d(keep, np.array([0]))
         trapped += temp*np.isin(labels, keep, invert=True)
+    logger.info("end of" + function_name)
     if return_mask:
         return trapped
     else:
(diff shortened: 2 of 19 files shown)


ma-sadeghi (changes requested):
@Daniel-olaO Thanks for the PR. Quick question: why do you use `inspect` to get function name? That's 2 extra lines of code per function (3 instead of 1), why not just hardcode the function name?

Also, please capitalize "start" and "end" for consistency. I think you also need to add a space between "start of" and function name.

jgostick:
Hi @ma-sadeghi, I suggested that the inspect module could be used to get the function name.  That way if we every change the function name or just cut-paste the code it will be self-healing.

ma-sadeghi:
Hey @jgostick, I see. What do you think about wrapping the "start" and "end" snippets in two private helpers, and just call them at the start/end of function calls, like `_log_enter_func()`, and `_log_exit_func`. There's a way to find the caller name:

```python
import inspect

def caller_function():
    # Get the frame of the caller function
    frame = inspect.currentframe().f_back

    # Get the name of the caller function
    caller_name = frame.f_code.co_name

    return caller_name

def test_function():
    caller = caller_function()
    print("Caller function:", caller)

test_function()
```

This would get rid of ~300-400 extra lines.

jgostick:
@Daniel-olaO I appreciate how thorough you were!  I didn't realize you were going to be so quick to implement this, and figured we'd have time to discuss and plan a bit.  

I think that we should be a bit more judicous with this feature an only put it in the "front-line" functions, and especially those that have the potential to be slow.  For instance, anything that starts with an underscore is a hidden function that is called by something else and anything that is `@jit` or `@njit` decorated is probably also not going to work with a logger inside.  Some for things that involve `dask`.  

Another thought...maybe we could jam all the inspect code into a single line?  That would make it harder to read, but easier to maintain.  Then we can cut-paste the code to new functions as needed for instance.

jgostick:
Not to make this overly complicated, but I wonder if this could be done as a decoarator?  I know we could do the 'start' but not sure about the 'end'. Like:

```python

@functimer
def function_name(args, kwargs):
    <code here>


def functimer():
   <code here>

```

I have not really dove into decorators so don't know the ins-and-outs, but this seems like a possibility.

Daniel-olaO:
Hey @jgostick, thank you for your input. I am considering implementing the solution suggested by @ma-sadeghi as it appears more apparent to me. What do you think about this?

ma-sadeghi:
@jgostick @Daniel-olaO I found an elegant workaround; the good thing about it is that it leaves all of our functions untouched, and requires a small modification to our modules' `__init__.py`:

Basically, what we need to do is to write a decorator (as @jgostick suggested) to be added to function entry/exit, and then apply it automatically to all our top-level functions as attribute. I've tested it on `generators` and it worked like a charm. We only need to add this snippet at the end of our `__init__.py` files:

```python
import sys
import logging
import functools

logger = logging.getLogger(__name__)

def log_entry_exit(func):
    @functools.wraps(func)
    def wrapper(*args, **kwargs):
        logger.info(f"Entering {func.__name__}")
        result = func(*args, **kwargs)
        logger.info(f"Exiting {func.__name__}")
        return result
    return wrapper

current_module = sys.modules[__name__]
for func_name, func in vars(current_module).copy().items():
    if callable(func):
        setattr(current_module, func_name, log_entry_exit(func))
```

@jgostick What do you think?

@Daniel-olaO In case we want to move forward, move `log_entry_exit` to `porespy.tools` so we don't need to redefine it for every module, then import it `from porespy.tools import log_entry_exit`.

image

jgostick:
@ma-sadeghi, wow that is slick.  Will it apply to EVERY function?  That could be a problem for functions that are called repeatedly. Do you think this will be a problem, and if so is there any way to specify which functions it can be applied to?

ma-sadeghi:
> @ma-sadeghi, wow that is slick. Will it apply to EVERY function? That could be a problem for functions that are called repeatedly. Do you think this will be a problem, and if so is there any way to specify which functions it can be applied to?

Good point. Currently, it'll apply to every accessible "top level" function within that module, both with or without underscore. I tried it on `networks` module, and here's the list of functions it gets applied to:

```python
add_boundary_regions
generate_voxel_image
label_phases
label_boundaries
map_to_regions
maximal_ball_wrapper
regions_to_network
snow2
_net_dict
_parse_pad_width
diffusive_size_factor_AI
create_model
find_conns
```

We could add a check like `not func_name.startswith("_")`, but there might still be functions without underscore that get called repeatedly (is this the case? If so, maybe it's a sign those should be underscored as well?).

If we want to have more control, we can add a list of functions we want to profile to `__init__.py`, and add a check inside the for loop: `if func_name in func_list`. It's a bit manual, but if there's no way to differentiate between "actual top level" functions vs. "helper" ones, this seems to be a good compromise.

Daniel-olaO:
Hey @ma-sadeghi, and @jgostick I have created the `log_entry_exit` in the tools module and added the import statement, just a quick question, how about calling the `log_entry_exit` for each function

ma-sadeghi:
> Hey @ma-sadeghi, and @jgostick I have created the `log_entry_exit` in the tools module and added the import statement, just a quick question, how about calling the `log_entry_exit` for each function

Great. You mean adding the decorator to each function? (before function header)

Daniel-olaO:
oh thanks, work on it

Daniel-olaO:
Hello everyone, I apologize for creating another PR with the number #839. I had to do this to resolve the logger issue. The reason for this is because I created a new branch from the dev branch to eliminate all the changes I made.

PR #841, opened 2023-05-24
Added logger decorator to all functions

### LINKED ISSUE #838: Add logger.info to beginning and end of each function

Daniel-olaO:
close #838

### DIFF
--- a/porespy/__init__.py
+++ b/porespy/__init__.py
@@ -24,6 +24,7 @@ from . import generators
 from . import simulations
 from . import visualization
 from . import io
+from porespy.tools import log_entry_exit
 # The dns module will be deprecated in V3, in favor of simulations
 from . import dns
 
--- a/porespy/filters/_funcs.py
+++ b/porespy/filters/_funcs.py
@@ -13,6 +13,7 @@ from porespy.tools import _check_for_singleton_axes
 from porespy.tools import get_border, subdivide, recombine, make_contiguous
 from porespy.tools import unpad, extract_subsection
 from porespy.tools import ps_disk, ps_ball, ps_round
+from porespy.tools import log_entry_exit
 from porespy import settings
 from porespy.tools import get_tqdm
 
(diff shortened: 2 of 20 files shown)


ma-sadeghi:
@Daniel-olaO I prefer the attribute approach. Adding decorators to every function, although is more explicit, but is less maintainable.

ma-sadeghi:
(especially since it's just for logging, if it were for something more substantial like just-in-time compilation, adding the decorator to function header would make sense since developers should really know what's going on, whereas for logging, the passive approach seems more sensible to me), but let's wait for @jgostick to weigh in before you make the changes to not waste your time. Also, I really appreciate your efforts :)

Daniel-olaO:
okay thanks @ma-sadeghi

jgostick:
There are pros/cons for both approaches.  The decorator is nice because it let's us explicitly pick and choose which functions to emit the logger message from.  I think @Daniel-olaO has gone a bit overboard by putting them on ALL the functions.  If we just do it on the really heavy functions, then I think the maintenence would not be too bad, AND the approach would be explicit.  On the otherhand, @ma-sadeghi's module attribute approach only requires code in ONE place, which is really tight.  The downside of this approach to me is that it also represents a maintenence issue in the future when we have to ask ourselves WTF these logger messages are coming from!

Daniel-olaO:
hI @jgostick, why not write a list of functions you want me to add the logger decorator

ma-sadeghi:
Another perspective: do we even need this? Profiling should be done by the user, and there are many great tools that exactly does this for us: scalene, memprofiler, etc. They provide per function runtime, memory usage, and even nice flame graphs.

jgostick:
I think this needs to be closed.  @Daniel-olaO I am sorry for discarding your hard work.

PR #871, opened 2023-07-17
Add ablog plugin

jgostick:


jgostick:
This is more trouble than it's worth...closing

PR #877, opened 2023-07-31
Compute tortuosity using domain decomposition and network modeling

rickyfann:


jgostick:
I'm going to merge this then fix the broken examples in a different PR

PR #922, opened 2024-03-10
Fixing pseudo packing functions

### LINKED ISSUE #575: Add `volume_fraction` as an optional argument to packing generators

### LINKED ISSUE #647: `pseudo_gravity` and `pseudo_electrostatic` packings should have `shape` or `im` as arguments
This is an api breaking change, so no rush.

### LINKED ISSUE #649: `pseudo_gravity_packing` does not work if some spheres already present
I was sure that this function could be used to add smaller spheres onto of exising ones, but as it is now this is giving an error.

### LINKED ISSUE #757: `clearance` arg in pseudo packings not working as it should
I am trying to make a image of single points (r=1) with a clearance of 10, but the pseudo_gravity_packing won't let me.  It complains that a negative particle size is not allowed or something like that. I guess the sanity check I put in there was not robust enough.

### LINKED ISSUE #805: Add a `packed_spheres` function to unify gravity and electrostatic packings
Both the gravity and electrostatic packing work by placing the spheres to be inserted as close to an 'attractor' as possible.  In the case of gravity, it's the bottom edge of the image, in the case of electrostatic it's the user-specified seed points. These could clearly be unified into a single function since the both do the same thing. To be consistent with the other functions this should be called `packed_spheres` (to match `random_spheres` and `lattice_spheres`).

### LINKED ISSUE #916: Redundant `edges` argument in `pseudo_gravity_packing`
The [documentation](https://porespy.org/modules/generated/generated/porespy.generators.pseudo_gravity_packing.html#porespy.generators.pseudo_gravity_packing) and the [function header](https://porespy.org/_modules/porespy/generators/_pseudo_packings.html#pseudo_gravity_packing) include the argument 'edges'. Changing this argument from default value does not change anything in the output.

Checking the source code revealed: it is not used within the function itself.

I would expect either an error/warning when changing the argument from default value or not having this argument at all.

jgostick:
Fixes #916
Fixes #757 
Fixes #649 
Fixes #647
Fixes #575 
Fixes #805

# Dont' Merge This PR Yet, Wait for V3

This PR contains reimplementations of the pseudo-packing functions. I rewrote them using a totally different logic so they don't output the same results as the previous ones (quantitatively different, qualitatively the same).  The new algorithms both use a queue based method and share a good chunk of code.  They seem to be about the same speed, but are much easier to understand now. They also work a lot better.  Not only do the original features all work now, but I have added several new args such as `phi` (solid volume fraction).

Anyway, because these functions return different results, it is technically mandatory to wait until releasing version 3 to merge this PR.


jgostick:
Got pseudo_gravity_packing working well, the clearance and mode args now work.  And adding new spheres to existing structures works...

Figure_1

jgostick:
Now that we have declared the `dev` branch to be "version 3" I think we can merge this PR.

PR #929, opened 2024-03-11
Refactor sem view

Anthero1:
Changed the way the SEM view is calculated. Now it will transpose the image to the appropriate axis and then probes into the image until it finds a solid voxel. The depth of that voxel is recorded for that x/y coordinate and saved in the resulting 2d image. This is also done with numba njit and prange to help speed up the process.


ma-sadeghi (dismissed):
Thanks for the PR. Could you please merge dev into your branch?

ma-sadeghi (commented):
Thanks, is this ready to merge now?

PR #958, opened 2024-05-08
Implement parallel extraction

Arenhart:
We have been working on the regions_to_network function, improving it's performance with Numba. We got an improvement of up to x60 faster network extraction when running in parallel. We also modified it to allow for anisotropic volumes.

While we tried to keep the results as close as possible to the original function, some differences were included, listed as following:

    Only valid for 3D volumes.
    Throat and perimeter calculations modified to allow anisotropic volumes (in a very simplified explanation, marching squares is applied to a flattened throat image).
    Distance local and global maxima is the average of all coordinates that have the highest distance transform value.
    Changed the marching cubes algorithm to an implementation that outputs the area and volume without the need of creating the mesh, "standard" accuracy was removed due to similar calculation time.

There are some minor fixes and modifications that got mixed in the branch, they could be moved to other issues/PRs.

Also, the branch shouldn't work unless Pyedt is available as a package. I will try to add it as soon as possible.


ma-sadeghi:
Thanks for the PR and the amazing speedup! To avoid git conflicts, please let us know once you're done with your changes, so we can review it.

jgostick:
Hi @Arenhart, I am not quite sure how to deal with PRs from other people's repos...like how can I pull your branch and make changes locally, etc. So, I am going to create a new `temp` branch, merge your code in there, then make all my changes on `temp`, and finally merge into `dev`.  I really appreciate your amazing contribution and I want it to wither on the vine, so I am going to forge ahead even if it's the wrong way to do it.

jgostick:
OK, I think I have figured out how to proceed, and also why all the conflicts.  Firstly, I have added the ltracegeo fork as an additional remote on my local PC, then pulled the branch.  Secondly, there are conflicts because @Arenhart edited some files that @ma-sadeghi moved to the `src` folder.  I will attempt to sort all this out, THEN push my updated branch to the `porespy` repo. I could probably try to push it to the `ltracegeo` repo, but I don't want to disturb their code.

PR #962, opened 2024-06-03
All distributions to uniform, fixes #961

### LINKED ISSUE #961: blobs porosity is off for porosity different than 0.5
Hi! 

When generating larger and larger images with `generators.blobs`, the porosity (when far away from 0.5) does not converge to the value given. 

```
import porespy as ps

l = 1000
blob1 = ps.generators.blobs([l,l], porosity=0.1, blobiness=l/50)
print(f'blob1 {l}x{l} mean = {blob1.mean()}')
l = 10000
blob2 = ps.generators.blobs([l,l], porosity=0.1, blobiness=l/50)
print(f'blob2 {l}x{l} mean = {blob2.mean()}')
```
```
blob1 1000x1000 mean = 0.101451
blob2 10000x10000 mean = 0.10158081
```

porespy version 2.4.1
python version 3.11.9
Mac OS

Thanks in advance, I think the issue is that after the gaussian filter in making the blobs image, the distribution is not normal.. It might be a bit slower but maybe have a all_to_uniform function that does not assume normality to change the distribution to uniform.

amirDahari1:
Changed the norm_to_uniform function to all_to_uniform function, to fix #961 since not all distributions that are needed for a specific porosity are normal, including the blobs distribution.


amirDahari1:
Added `porosity` argument to `fractal_noise` generator to fix #963. I'm not sure if to keep the `uniform` argument or not, since `porosity` already needs the transition to uniform, but thought to keep it for the case when the user prefers the uniform distribution without thresholding.

jgostick:
I am going to merge this PR, then sort out the broken examples afterwards.  Thank you very much for this!.

PR #980, opened 2024-07-29
Avoid overflow errors with _get_radial_sum

cmaloney111:
The `_get_radial_sum` function currently overflows for large autocorrelation values. This can lead to negative values for the probability in the `two_point_correlation` function.

**Problem**
When using the two_point_correlation function with large autocorrelation values, I encountered overflow errors:

- When using porespy through Anaconda, the overflow results in negative probability values (see attached screenshot).
- On my local machine (where I copied the `two_point_correlation` code), it causes an OverflowError with the message "Python int too large to convert to C long."

**Solution**
To fix this issue, I:

- Specified the dtype of radial_sum as float64 to handle large values without overflow.
- Added a check to avoid division by zero using if np.any(mask).

Two Point Correlation


jgostick:
Thanks for this.  This reminds me of another bug I fixed recently, where something `np.sum(im)` would use the `dtype` of `im` when doing the sum.  I fixed it by specifying the dtype in the sum function, similar to what you did here.  Thanks for catching this.

cmaloney111:
No problem!

PR #991, opened 2024-10-15
Fixed edges of blobs to maintain uniform density

### LINKED ISSUE #981: Different phase fraction on edges of blobs output
On the edges of blobs output, there is a higher phase fraction then the inside of the output.

Here you can see the phase fraction for every slice along the first dimension in a 500x500x500 blobs output with phase fraction 0.1:

porespy_issue

jgostick:
Fixes #981 

- Changed `mode` in `gaussian_filter` to `mirror` by default
- Added `periodic` argument to `blobs` that if `True` will use `wrap` mode in `gaussian_filter` to create a periodic image
- Updated `blobs` example to describe `periodic` as well as `seed` arguments.


jgostick:
We are going to do this a different way, as discussed in #1024

PR #996, opened 2024-10-28
Fix typo

habi:


jgostick:
I do appreciate the PR, but triggering 20+ builds and test suites to change one single character in the readme file seems senseless.  We are in the middle of moving to V3 so I will make a mental note to spell check things. I will close this now, sorry.

PR #1033, opened 2025-02-19
Updating the documentation in preparation for V3

jgostick:


ma-sadeghi:
@mkaguer Could you please update `magnet_in_detail` notebook? It's currently failing because it uses some nonexistent functions, like `insert_pore_bodies`, etc. Thanks!

jgostick:
This PR is getting really big and sprawling.  Should we merge it and start another 'final' one to do all the remaining tidyups?

PR #1069, opened 2025-04-22
Adding a "C" function for measuring randomly directed chords

jgostick:
The chord length function is useful, but the "C" compilation machinery/setup is the most interesting development here as this charts new territory for us.  

This PR includes
- Some C-source files
- Python wrapper to call chord length function (currently in beta module)
- Updated actions to compile with meson
- Action uses Pypa approved action to build wheels for all platforms
- Also added a new action to publish to Pypi, using Pypa approved action instead of hatch


guoalex1:
Some changes to build correctly:
Set macOS version environment variable for cibw
Add back Linux, updated pyproject.toml to correctly include the shared library file in the wheel

Everything can now build correctly, we might need a way to change the version number automatically or just update it manually to upload to TestPyPI.

ma-sadeghi:
We should probably add a unit test too.

jgostick:
@guoalex1, thanks very much for fixing this! I was started to play LLM-roulette to see if claude could figure out how to fix it...it was not going well.  

@ma-sadeghi, this function needs a bit of work still before we can finalize tests.  I do wonder if the code coverage will work on it though, or if we'll have to comment it out like with the numba functions?

ma-sadeghi:
@jgostick Agreed on the final unit test, but I meant just adding a dummy test to just call this function to check if it runs without problems (outputs something without errors).

For the automatic versioning, we can use something similar to the PyPI workflow. I can look into it this weekend if you haven't already by then.

PR #1072, opened 2025-04-26
Resolve deprecated usage of bitwise operators on Python Boolean types

emmanuel-ferdman:
# PR Summary
This small PR resolves the deprecation warnings coming from the bitwise operators on Python Boolean types:
```python
/Users/runner/work/porespy/porespy/test/unit/test_filters.py:28: DeprecationWarning: Bitwise inversion '~' on bool is deprecated and will be removed in Python 3.16. This returns the bitwise inversion of the underlying int object and is usually not what you expect from negating a bool. Use the 'not' operator for boolean negation or ~int(x) if you really want the bitwise inversion of the underlying int.
```
See [CI logs](https://github.com/PMEAL/porespy/actions/runs/14678020927/job/41197103505#step:7:343) for more information.

### DIFF
--- a/test/unit/test_filters.py
+++ b/test/unit/test_filters.py
@@ -25,7 +25,7 @@ class FilterTest():
     def test_im_in_not_im_out(self):
         im = self.im[:, :, 50]
         for item in ps.filters.__dir__():
-            if ~item.startswith('__'):
+            if not item.startswith('__'):
                 temp = getattr(ps.filters, item)
                 assert temp is not im
 


jgostick:
I appreciate the PR @emmanuel-ferdman, but I'm a bit surprized this is only happen in one place.  I just searched the code base for `if ~` and found this usage is also on:

line 389 of _injection.py
line 81 of _ibip_gpu.py

I wonder what other search strings would turn up something?

jgostick:
Just searching for `~` returns 217 usages. We use this regularly to invert numpy boolean images (which I *really* hope is not going to be deprecated!).

PR #1100, opened 2025-07-10
Adding new implementations of the `local_thickness` filter plus related changes

### LINKED ISSUE #1099: Move `porosimetry` to `simulations` and rename `hg_porosimetry`?
The `porosimetry` function is literally the oldest function in Porespy.  I used this function a lot to compare to real MIP data, so created Porespy to hold this function and share it.  I think the time may have come to change things.  We now have the `simulations` module, which contains the reference implementations of `drainage` and `imbibition`.  The `porosimetry` function is now obsolete since people should use `drainage` with the capillary transform representing Hg properties.  

In parallel to this, the `local_thickness` filter actually calls the `porosimetry` function.  I have been dabbling with different implementation of the local thickness filter, inspired by the review of [Dahl and Dahl](https://openaccess.thecvf.com/content/CVPR2023W/CVMI/html/Dahl_Fast_Local_Thickness_CVPRW_2023_paper.html), which has been published as a [github package too](https://github.com/vedranaa/local-thickness).  

So here is what I am thinking:
1. Remove `porosimetry` from `filters`, moving all the logic from it to the `local_thickness` function.
2. Add logic to the `local_thickness` function to perform it in different ways, such as the method of Dahl and Dahl, but also my attempt to implementing using the same logic as MorphoJ (the imageJ plugin).  Basically, we'd promote the `local_thickness` function to a first-class citizen.
3. In the `simulations` folder we can create a function called `hg_porosimetry` which is a wrapper for drainage and imbibition, where we create the capillary transform using the properties of Hg, perform a drainage with no trapping and all sides as inlets, then also perform an imbibition *with trapping*.  The returned `Results` object would have both intrusion and extrusion data on it, just like an MIP experiment.  This would provide people with a simple, convenient way to compare with MIP, but would use the official `drainage` and `imbibition` functions.  

Obviously, this is a breaking API change, so we'd need to include this in V3.

jgostick:
Closes #1099

- `local_thickness` is now fully implemented rather than calling `porosimetry`
- There are now several implementations of `local_thickness` like `local_thickness_dt`, `local_thickness_conv`, etc.  Each of these can be called directly, or by specifying a `method` in the generic `local_thickness` function.
- Added an attempt to perform local thickness using the logic in ImageJ. This is not as fast as the `dt` method though.
- Added a brute-force method which provides the *exact* local thickness for reference. Note that the ImageJ method matches the brute-force method with about 5x faster performance. It's possible to do a lower accuracy version of the ImageJ method with is another 2x faster, but is no longer an exact match.
- Rewrote the `porosimetry` method in the `filters` module to use the `drainage_xyz` functions in the `simulations` module. 
- Added a `hg_porosimetry` method to the `simulations` module which does intrusion and extrusion of Hg using predefined settings, which should be convenient.
- All `drainage_xyz` and `imbibition_xyz` methods now accept `dt` as an optional argument
- Changes the way `steps` are treated in the `drainage_xyz` and `imbibition_xyz` functions, and they now all called `parse_steps` function in the `tools` module


jgostick:
Closes #1099

PR #1127, opened 2025-10-17
Fix typos

habi:


jgostick:
Hi @habi, thanks for this!  I have now got spell checker on my IDE so hopefully there will be less typos in the future :-)

PR #1141, opened 2025-12-02
Drop Python 3.10, and begin supporting Python 3.13

ma-sadeghi:
Per [NumPy version compatibility table](https://numpy.org/neps/nep-0029-deprecation_policy.html), Python 3.11+ is encouraged as of April 2025.

It'd be nice to support 3.14 too, but `triangle` is not yet supported.

@jgostick I also moved our optional dependencies into the extras section of pyproject.toml, so they’re no longer installed by default. My thinking was that we already catch `ModuleNotFoundError` for all of them. If you had something else in mind, let me know.


ma-sadeghi:
@jgostick After updating `uv.lock`, the following unit tests fail:

```python
FAILED test/unit/test_simulations_ibop.py::IBOPTest::test_drainage_equals_drainage_dt_smooth - AssertionError
FAILED test/unit/test_simulations_ibop.py::IBOPTest::test_drainage_equals_drainage_dt_not_smooth - AssertionError
FAILED test/unit/test_simulations_ibop.py::IBOPTest::test_imbibition_equals_imbibition_dt_smooth - AssertionError
FAILED test/unit/test_simulations_ibop.py::IBOPTest::test_imbibition_equals_imbibition_dt_not_smooth - AssertionError
```

Could you please check these four unit tests and see if they work as expected?

jgostick:
The failing tests all belong to IBOP. I will dig into it in a few days.

PR #1157, opened 2026-04-23
Touching up stl funcs

jgostick:
Since the `from_stl` function just received some love, I realized the `to_stl` function was no longer up to par. Also, I am hoping that an STL which just includes the surface pores might rendering reasonably efficiently.


jgostick:
- Big upgrade to `to_stl` #enh

PR #1162, opened 2026-04-26
Fix IBOP equivalence tests by computing `pc = 2/dt` at float64 precision

### LINKED ISSUE #1145: Fix skipped IBOP equivalence tests
@jgostick After updating `uv.lock`, the following unit tests fail:

```python
FAILED test/unit/test_simulations_ibop.py::IBOPTest::test_drainage_equals_drainage_dt_smooth - AssertionError
FAILED test/unit/test_simulations_ibop.py::IBOPTest::test_drainage_equals_drainage_dt_not_smooth - AssertionError
FAILED test/unit/test_simulations_ibop.py::IBOPTest::test_imbibition_equals_imbibition_dt_smooth - AssertionError
FAILED test/unit/test_simulations_ibop.py::IBOPTest::test_imbibition_equals_imbibition_dt_not_smooth - AssertionError
```

Could you please check these four unit tests and see if they work as expected?

_Originally posted by @ma-sadeghi in https://github.com/PMEAL/porespy/issues/1141#issuecomment-3600416323_

ma-sadeghi:
Fixes #1145.

`edt` returns float32, so the default `pc = 2/dt` inside `drainage` and `imbibition` is also float32. For voxels at integer `dt` boundaries (e.g. `dt = 3.0`), the float32 result of `2/3.0` is `0.6666666865` — about `1e-8` above the float64 threshold `2/3 ≈ 0.6666666667` produced by the steps array. Those voxels then fail `pc <= P` and end up one step behind the integer-`dt` branch, which is exactly what the four `*_equals_*_dt_*` IBOP tests were catching before they got skipped.

Cast `dt` to float64 before computing `pc` inside the two functions, and update the drainage tests (which pass `pc` explicitly) to do the same. Imbibition tests don't pass `pc`, so the function fix alone is enough for them.

The four skipped tests now pass with `diff_size = 0` and `diff_seq = 0` for the `smooth=True` cases, beating the original `< 5` tolerance jgostick set on the `smooth=False` ones (3 stray pixels remain there, attributed to the discretization gap between `_insert_disks_at_points_parallel` and an `edt`-based dilation).

Full unit suite still green: 329 passed, 9 unrelated skips.

### DIFF
--- a/src/porespy/simulations/_drainage.py
+++ b/src/porespy/simulations/_drainage.py
@@ -569,7 +569,10 @@ def drainage(
         dt = edt(im)
 
     if pc is None:
-        pc = 2.0/dt
+        # Cast `dt` to float64 to avoid precision loss in `2.0/dt`. With float32
+        # `dt`, voxels at integer `dt` boundaries miss the `pc <= P` threshold
+        # by ~1e-8, putting them out of sync with integer-`dt` step sequences.
+        pc = 2.0/dt.astype(np.float64)
     pc[~im] = 0  # Remove any infs or nans from pc computation
 
     if isinstance(steps, int):  # Use values in pc for invasion steps
--- a/src/porespy/simulations/_imbibition.py
+++ b/src/porespy/simulations/_imbibition.py
@@ -579,7 +579,10 @@ def imbibition(
         dt = edt(im)
 
     if pc is None:
-        pc = 2/dt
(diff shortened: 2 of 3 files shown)


ma-sadeghi:
Pushed a small fix to drop the unused `pytest` import. The remaining `test_bond_number` failure on py3.11 ubuntu is the same flake fixed in #1161 — it'll clear once that lands and this branch is rebased.

PR #1164, opened 2026-04-26
Add `return_indices` to `local_thickness` and `porosimetry` for compact output

### LINKED ISSUE #840: Lower memory footprint for `porosimetry` and `local_thickness`
Hi,

Thank you for all the hard work on this package. This is a treasure trove!

I have to use `local_thickness` of huge tomography dataset (1500 x 1500 x 1200 voxels) and I had a hard time to do so even on a workstation with 64 Gb of RAM. Therefore, I looked into the code, and found a way to reduced the memory footprint. I implemented it on my side, it seems to work, so I though I could share it here.

At the moment, `porosimetry` returns an array of pore sizes, one float per voxel of the original image. In my case, this is 21Gb in memory just for the output. However, the values stored in this output array are discreet. By construction, only some values of sizes are tested. The default if 25 sizes, spaced logarithmically.

What I propose, is to return two arrays:

- sizes: A 1D array of floats containing all the tested sizes, typically 25 values
- indices: An array of unsigned ints that contains **the index of** the pore size for each pixel.

If we allow at most 255 possible discreet sizes, the type of the second array can be uint8, taking less than 3Gb in my case. If we allow all 65535 values that fit in uint16 (crazy use case), the second array is still less than 6 Gb in my case. This would leave much most space to perform the rest of the algorithm (Euclidian distance transform, fft dilation, etc.).

If one wants the original array of floats, it is as simple as sizes[indices]. However, if one wants the pore size distribution, it is faster to compute the histogram of the indices and to plot it function of sizes.

In practice, here are the changes in the code of `porespy.filters._funcs.porosimetry`

Change
```python:
imresults = np.zeros(np.shape(im))
for r in tqdm(sizes, **settings.tqdm):
    imtemp = dt >= r
    ...
    imresults[(imresults == 0) * imtemp] = r
```
to
```python:
imresults = np.zeros(np.shape(im), np.uint8)
for ir in tqdm(range(len(sizes)), **settings.tqdm):
    r = sizes[ir]
    imtemp = dt >= r
    ...
    imresults[(imresults == 0) * imtemp] = ir+1
```
(To be done 3 times for the 3 possible modes)

And at the end of the function `return np.concatenate(([0], sizes)), imresults` instead of `return imresults`.

(You may  notice that the indices are shifted by +1 in order to accommodate 0 as the index of the foreground).

ma-sadeghi:
Closes #840.

The default float64 output of `local_thickness`/`porosimetry` becomes the peak-memory bottleneck on large tomograms (~21 GB for a 1500x1500x1200 volume in OP's case). With `return_indices=True` these now return `(sizes, indices)` where `indices` is `uint8`/`uint16` and `sizes[indices]` recovers the float result. Output memory drops 8x or more.

To save memory inside `porosimetry` too (not just at the API boundary), `drainage_dt` gained a matching `return_indices` flag that skips allocating the float `im_size` array and writes the step index into a compact `im_seq`. The mode is opt-in and not compatible with `outlets`, since trapping post-processing relies on `im_size`.

Only `method='dt'` is supported for now; the other methods raise `NotImplementedError`. Easy to extend to `conv`/`bf` in a follow-up if needed.

```python
# Before: 21 GB float64 output
lt = ps.filters.local_thickness(im, sizes=25)

# After: ~3 GB uint8 output, sizes[indices] recovers the float
sizes, indices = ps.filters.local_thickness(im, sizes=25, return_indices=True)
```

Default behaviour is unchanged. Two new tests added; full unit suite green.

### DIFF
--- a/src/porespy/filters/_lt_methods.py
+++ b/src/porespy/filters/_lt_methods.py
@@ -33,6 +33,15 @@ __all__ = [
 ]
 
 
+def _index_dtype(n):
+    """Smallest unsigned int dtype that holds values `0..n` inclusive."""
+    if n <= np.iinfo(np.uint8).max:
+        return np.uint8
+    if n <= np.iinfo(np.uint16).max:
+        return np.uint16
+    return np.uint32
+
+
 def porosimetry(
     im: npt.NDArray,
     dt: npt.NDArray = None,
@@ -40,6 +49,7 @@ def porosimetry(
     sizes: int = None,
     method: Literal['dsi', 'fft', 'dt'] = 'dt',
     smooth: bool = True,
+    return_indices: bool = False,
 ):
     r"""
     Each location is assigned the radius of the largest sphere that can reach it
@@ -80,12 +90,19 @@ def porosimetry(
     smooth : bool, optional
         Indicates if protrusions should be removed from the faces of the spheres
         or not. Default is `True`.
+    return_indices : bool, optional
+        If `True`, return `(sizes, indices)` instead of a float image. `sizes` is
+        a 1D array of applied radii with a leading `0` for unfilled voxels, and
+        `indices` is an unsigned-int image such that `sizes[indices]` reproduces
+        the float result. Avoids allocating a full float64 image, which matters
+        for large tomograms. Only supported for `method='dt'`. Default is `False`.
 
     Returns
     -------
     sizes : ndarray
         In image with each voxel value indicating the largest overlapping sphere
-        which can reach it from the given inlets.
+        which can reach it from the given inlets. If `return_indices` is `True`,
+        returns `(sizes, indices)` instead.
 
     See Also
     --------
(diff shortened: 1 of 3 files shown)


ma-sadeghi:
The `test_bond_number` failure on py3.13 ubuntu is the same flake fixed in #1161 — it'll clear once that lands and this branch is rebased. Everything else is green.

jgostick:
I wonder if this is the right way to fix this?  The reality is that all spheres drawn in a digital image have integer radii, so writing fractional radii is actually an artifact of the way the sizes are selected. If sizes is given as N, then it pulls out N sizes between 1 and max(dt).  We could change the default behavior of the size generator, and then make the output (im_sizes) respect the dtype of the sizes? (i.e. int32)

jgostick:
I have added this feature to the new #1192 work, so closing this

PR #1165, opened 2026-04-26
Add `rescale_network` and `param.*` metadata

### LINKED ISSUE #614: Allow rescaling SNOW based on voxel size
Currently, voxel size is passed as an argument to SNOW, meaning that once the network is extracted, it's not possible to easily rescale the output based on a different voxel size. It'd be handy if there was an option to rescale the network afterwards. This is specially useful for large extractions that might take days to complete.

ma-sadeghi:
Closes #614.

Network extractions can take days for large images, but until now changing voxel size required re-running the whole thing. This adds:

1. `param.voxel_size` and `param.ndim` written into the network dict by `regions_to_network`, `regions_to_network_parallel`, and `magnet`.
2. `ps.networks.rescale_network(network, voxel_size)` which reads those, computes the scale factor, and rescales all length / area / volume fields. Output matches a fresh extraction at the new voxel size.

The function is name-driven: it inspects the part after the last `.` (e.g. `pore.inscribed_diameter` -> `inscribed_diameter`) and looks up an exponent (1 for length-like, 2 for area, ndim for volume, 1 for size-factor-like). Anything it doesn't recognize is left alone.

`param.*` is the OpenPNM-blessed namespace for scalar metadata. PMEAL/OpenPNM#3046 fixes `network_from_porespy` so these keys round-trip cleanly through OpenPNM project save/load. Without that fix, the keys still work in-memory and through plain pickle, just not through `.pnm` save/load.

Anisotropic voxel sizes (`regions_to_network_parallel` accepts a tuple) raise `NotImplementedError` from `rescale_network` if the components differ. Same-on-all-axes tuples are treated as scalar. `rescale_network` returns a new dict; the input is not mutated.


jgostick:
This seems like a great addition, and I like the implementation.

Issue #950, opened 2024-03-27
`overlay` does not work in 2D


jgostick:

jgostick:
This is captured in #951 so closing

PR #1168, opened 2026-04-27
Replace `overlay` with `insert_shape_at_points`

### LINKED ISSUE #951: Replace `overlay` with `insert_shape_at_point`
The overlay function should accept a mode, which can be one of `'add', 'preserve', 'overwrite'`.  It's currently only doing `'add'`.

ma-sadeghi:
Closes #951. Also fixes the 2D bug from #950 as a side effect.

Per @jgostick's comment in the issue thread, the "stamp this shape at these coordinates" inner loop was duplicated across the disk variants. Pulled it out into three numba primitives that take an `element` argument: `_insert_shape_at_point`, `_insert_shape_at_points`, `_insert_shape_at_points_parallel`. The existing `_insert_disk_at_point[s][_parallel]` collapse to thin wrappers that build the disk/ball stencil and delegate. Disk-path behaviour is bit-for-bit identical to dev across all boundary, smooth/non-smooth, and overwrite/preserve cases.

Public `insert_shape_at_points` exposes the same primitive directly. Accepts a single coordinate or an `(ndim, npts)` batch. Modes: `'preserve'`, `'overwrite'`, `'add'` (with `'overlay'` as an alias).

`tools.overlay` is now a deprecation shim that routes through `insert_shape_at_points` with `mode='add'`. Keeps existing callers working, fixes the 2D failure mode from #950, and emits a `DeprecationWarning` pointing at the replacement. The two internal callers in `networks/_funcs.py::generate_voxel_image` switch to `insert_shape_at_points` directly. The old `try/except ValueError` fallback for edge pores is gone because the new primitive clips out-of-bounds writes natively.

Plural-radii variants `_insert_disks_at_points*` are untouched. They take per-point radii so they don't fit the single-element shape primitive cleanly. Could be folded into a `_insert_shapes_at_points` later.


jgostick:
There is a feature I would like to add to these functions: at present there is a flag for 'overwrite' or not.  I would like to change this to a 'mode' where options are 'overwrite' and 'add' (plus any future behaviors).  The reason is that adding 1's to voxel repeatedly (`mode='add'`) lets you find overlaps, which I've wanted to do a few times lately.  The default would be `mode=None`, which would behave like `overwrite=False`.

PR #1174, opened 2026-04-28
Fix two_point_correlation radial average (#966)

### LINKED ISSUE #966: `two_point_correlation_fft` is only calculated for one orthant
`two_point_correlation` calculates the average of the TPC of only one orthant (quadrant 2D/ octant 3D), which might be different than the average over all vector directions.

ma-sadeghi:
Fixes #966.

The FFT TPC was averaging over only one orthant, but digging in, that's one of three compounding bugs in the radial profile:

1. `rfftn` followed by `ifftshift` on the half-Hermitian spectrum, which has asymmetric shape `(N, ..., N//2+1)` on the last axis. The `fftshift`/`ifftshift` dance on it is undefined and the autocorrelation came out biased along that axis.
2. No zero-padding, so the FFT computed the *periodic* autocorrelation. Non-periodic images get near-edge voxels wrapped into spurious far-lag pairs.
3. `_get_radial_sum` used a mask `(dt > r - bin_size) & (dt <= r)`, which represents `(bins[i-1], bins[i]]`, but `_parse_histogram` labels bin centres assuming `[bins[i], bins[i+1])`. Reported values were one bin off.

A `dtype=np.int64` cast inside the sum was also truncating fractional autocorrelation to 0, working by accident because the unnormalized autocorrelation is near-integer for boolean inputs.

Fix: zero-pad to 2N, normalize each lag by its voxel-pair count so `S2(r) == pf` at lag 0, use `pf` instead of `np.max(autocorr)` for the radial-profile normalization (the unbiased estimator can spike at noisy edge lags), switch bin masks to `[bins[i], bins[i+1])` so values match labels, and floor-divide the centre index so odd-shaped arrays put zero-lag on a grid point.

Verified against an explicit angular average of `S2` (FFT agrees to ~0.005 across r vs ~0.1 before) and against `two_point_correlation_bf`.

`test_tpcf_fft_3d_scaled` used `lattice_spheres` to check decay to `phi**2`, but lattice images have an oscillatory TPC that doesn't actually decay at finite r. It was passing by accident on the buggy curve. Switched to a blobs image, which does decay. Added `test_tpcf_fft_averages_over_all_orthants` as a direct regression for the orthant issue.


jgostick:
Wow, that is big improvement.  I'm glad that comparing to the bf method came in handy.  We don't import that function because it's painfully slow, but it turned out to be useful.

PR #1175, opened 2026-04-28
Add `satn_to_time` filter (#940)

### LINKED ISSUE #940: Add a `satn_to_time` function
This would be useful for making movies which are properly scaled in time.

ma-sadeghi:
Closes #940.

Replaces the commented-out skeleton in `_size_seq_satn.py` with a working `satn_to_time(satn, im, flow_rate, voxel_size=1.0, mode='drainage')`. For constant-flow-rate injection the cumulative invading volume is just `flow_rate * t`, so the time at which a voxel was invaded is `satn * V_void / flow_rate`.

The commented out code:

> ```python
> # Seems to color in wrong order for imbibition
> # def satn_to_time(im, satn, flow_rate, voxel_size, mode='imbibition'):
> #     f = 1 if mode == 'imbibition' else 0
> #     bins = np.digitize(satn.flatten()-f, bins=np.unique(satn-f), right=True)
> #     counts = np.cumsum(np.bincount(bins))
> #     time = counts*(voxel_size**im.ndim)/flow_rate
> #     time_map = np.reshape(time[bins], im.shape)
> #     plt.imshow(time_map/im)
> ```

seems to have been a saturation-convention mismatch. `seq_to_satn` returns the *invading*-phase saturation in drainage but the *defending*-phase saturation in imbibition, so the imbibition formula needs `(1 - satn)` instead. Verified that both modes give identical per-voxel times on the same simulation.

Solid voxels stay at 0 and trapped voxels (`satn == -1`) propagate as -1 in the output, matching the rest of the `*_to_*` family.

Tests cover both modes, voxel-size scaling, trapping propagation, solid handling, and the unknown-mode error.

Worth a follow-up reference notebook in `examples/filters/reference/satn_to_time.ipynb` since the docstring already points at the URL — happy to do that in a separate PR once this lands.


jgostick:
This feature is basically just a unit conversion, incorporating the voxel size and given flow rate.  It is pretty trivial, but I suppose it's marginally useful.

PR #1176, opened 2026-04-28
Wire `regions_to_network_parallel` into `snow2`

### LINKED ISSUE #1058: Incorporate `regions_to_network_parallel` into `snow2`
The `regions_to_network_parallel` function is available in the `networks` module but is not used in the `snow2` function.  We should consider adding an flag, or perhaps a keyword in the `parallel` argument that can enable this.

ma-sadeghi:
Closes #1058.

Adds a new `parallel_extraction_kw` argument to `snow2`. When `None` (default), the standard `regions_to_network` is used; passing a dict switches to the numba-threaded `regions_to_network_parallel`, with any options (e.g. `threads`) forwarded as kwargs.

Kept separate from `parallel_kw` since that one is scoped to chunked watershed partitioning — bundling the two would conflate two different parallelism mechanisms (per @mkaguer's note on the issue). Raises on 2D images since the parallel extractor is 3D-only, and widens scalar `voxel_size` to a tuple before forwarding.

Two new tests: one asserts the 2D guard, the other (skipped when `pyedt` is missing) compares the network shapes from the serial and parallel extractors on a 3D blob.


jgostick:
I am not sure I love this proliferation of keywords args.  I am thinking that maybe a new function should be added: `snow_parallel`, which can support both parallelization tactics?

jgostick:
I have created a new 'api', which also support backward compatibility.  You do `parallel_kw={'watershed': {kws here}, 'extraction': {kws here}}`. If you do the old way of `parallel_ks={kws here}` is also works to parallelize the watershed step, though eventually we'll want to deprecate that eventually.

PR #1177, opened 2026-04-28
Promote `flux` and `tau_from_cmap` from `beta` to `simulations`

### LINKED ISSUE #1023: Add function for converting concentration map to tortuosity
This was written in Julia, but I think a python version is kicking around too?  

@ma-sadeghi, do you have this handy?

ma-sadeghi:
Closes #1023.

Moves `flux` and `tau_from_cmap` from `porespy.beta._dns_tools` into `porespy.simulations._dns_tools`, alongside `tortuosity_fd`. Pure rename — no code changes — plus the import plumbing in `simulations/__init__.py` and `beta/__init__.py` and the test imports.

On the edge-case bug I asked about in the issue thread: I went looking for it and couldn't reproduce one. I swept several hundred combinations of `ndim ∈ {2, 3}`, varied shapes (cubic, anisotropic, long-thin, short-fat down to `L=2`), porosities `0.4–0.85`, multiple seeds, all axes, plus targeted constructions (single-column channels, partially-blocked inlet faces, multiple disjoint percolating clusters, last-layer solids that exercise the `_fix_gradient_outlet` path). In every case `tau_from_cmap(im_conc, im, axis)` agreed with `tortuosity_fd(im, axis).tortuosity` at solver-noise level (worst rel-diff ~1.4e-5, typically 1e-6 or tighter). The boundary fix at the outlet is approximate but safe under `tortuosity_fd`'s outputs because solid voxels and Dirichlet-zero outlet voxels both have `c=0`, which kills the only term that could disagree.

Three new tests are added to capture the cases I checked manually so future readers can re-audit: low-porosity 2D blobs, an inlet face with a solid half, and an anisotropic 3D shape across all three axes.

Heads-up for #1170 if it lands second: it imports `from porespy.beta import flux`; that path is gone and needs to become `from porespy.simulations import flux`.


jgostick:
@ma-sadeghi I had to reduce the rtol on your 3 tests to 1e-3 from 1e-4 and 1e-5.  Maybe you can rerun your checks?

